PR #1, opened 2015-11-03
Make Qt project compile and work as expected

Nitnelav:
the project was not compiling... this just fixes it. No additionnal functionnality

### DIFF
--- a/qt/pubnub_qt.cpp
+++ b/qt/pubnub_qt.cpp
@@ -6,6 +6,7 @@ extern "C" {
 }
 
 #include <QtNetwork>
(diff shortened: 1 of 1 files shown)


vveljko:
Hi, sorry about this, we were aware of it and the fix just slipped being committed because it was done as part of a larger effort. But, we'll provide a "separate" fix for it shortly and you can try it out, and if you still have issues after that, let us know. Thanks!

vveljko:
Please check the latest commit, it has the fix we mentioned earlier (and a few others, actually).
Thanks again.

PR #2, opened 2015-12-04
Fix memory leak

walterbrebels:
Free the certificate after using it.


vveljko:
Thank your for this pull request.

PR #3, opened 2015-12-07
Cannot start new openssl transaction after previous transaction was cancelled

walterbrebels:
When an openssl transaction was cancelled it was not possible to start a new transaction.

It would perhaps be a good idea to write a test for this, but the tests are currently failing on my system because quota are exceeded with the used account.


vveljko:
It's actually pretty much the same as is done in "generic sockets" PAL, it just didn't get updated for OpenSSL for some reason, so we're pretty sure it's fine.

Thank you for this pull request.

PR #4, opened 2015-12-14
OpenSSL: Set timeout

walterbrebels:
Use timeout on openssl connection to ensure BIO_read does not block infinitely


vveljko:
This was actually on our "to-do" list. :)

Also, this:

    if (PUBNUB_BLOCKING_IO_SETTABLE && pb->options.use_blocking_io) {
        return PNR_IO_ERROR;
    }

should be handled a little differently, but, it's OK for now.

In any case, thank you for this pull request.

PR #6, opened 2016-04-13
Minor fixes/improvements

sjaeckel:
- `core/pubnub_ssl.c` was not compiling
- we should be able to set the buffer sizes from outside the project


vveljko:
Thank you for this pull request.

MaxPresman:
@sjaeckel Thanks Steffen!

PR #11, opened 2017-01-29
Fixed reference to SSL_CTX member of pubnub_pal

rmadair:
- The SSL_CTX member of pubnub_pal is called ctx instead of sslCtx.
- Semicolon was needed
- wincrypt needs windows.h


vveljko:
It seems wrong version of this file was committed, sorry about that.

Thank you for this PR.

PR #12, opened 2017-04-03
pubnub::futres must unregister its callback upon destruction.

chriskush:
Neglecting to do so can cause other threads to crash; for example, if they call pubnub::context::cancel() while no transactions are in progress on that context.


vveljko:
Thank you for this PR.

PR #29, opened 2017-09-24
Fixed time-parameter for pthread_cond_timedwait() function.

MacGritsch:
Proofed for macOS/Darwin but not for other Posix Systems.

This fix should solve the following issue:
If you try to start listening to a pubnub channel when no internet-connection is available, CPU usage goes to 100% and even does not go down when a connection is established.

Digged into that problem on our macOS Systems and found the problem:
The function pthread_cond_timedwait() expects a time-interval since january 1st 1970 as "abstime"-parameter, but currently the time-interval since the last reboot is passed. Therefore pthread_cond_timedwait() exits immediately and the endless-loop continues.
The macOS man-page for pthread_cond_timedwait() is really clear and says that gettimeofday() (CLOCK_REALTIME) is used to obtain the time that is compared to "abstime"-parameter.
The man-pages for other POSIX-systems are not that clear and they only speak of "absolute time". It also seems to be the CLOCK_REALTIME (and not CLOCK_MONOTONIC as pubnub code is currently using). I have checked this with the function pthread_condattr_getclock() on different posix-systems (online c-compilers).


vveljko:
@MacGritsch Hi and thank you very much for your PR and comments. It does point out an error in our code which is tricky to find and has various negative consequences.

There a few issues with your PR:
- It's "against" the `master` branch, but, it should be "against" the `develop` branch. We have some internal policies about merging to `master`, wouldn't make much sense to bother you with them.
- On older MacOSX versions, there is no `clock_gettime()`, so, removal of our "replacement" will break the build on older versions.
- The time that was read was really meant to be `CLOCK_MONOTONIC` (for the timers), so, this change will break the timers support. Thus, we actually need to add a new time variable, reading the `CLOCK_REALTIME` for the purposes of `pthread_cond_timedwait`.

Let us know if you wish to address these issues in a new PR, or, if you don't want to do that, we can use your valuable insights in making the fix ourselves.

MacGritsch:
Hi @vveljko,

- ok, I can create a PR against `develop`

- ok, can change to gettimeofday for Apple products using code like this:
`void clock_get_realtime(struct timespec *ts)
{
#if defined(__APPLE__)
	
	struct timeval tv;
	gettimeofday(&tv, NULL);
	TIMEVAL_TO_TIMESPEC(&tv, ts);
	
#else
	
	clock_gettime(CLOCK_REALTIME, ts);
	
#endif
}`

- I have grep'ed the whole code and have not found usage of the `monotonic_clock_get_time()` function in other code.

MacGritsch:
This PR can be closed as I created a new one agains `develop` with changes for older OS X versions:

https://github.com/pubnub/c-core/pull/29

PR #30, opened 2017-09-26
Fixed time-parameter for pthread_cond_timedwait() function.

MacGritsch:
For details see commit-comments.


posix/pubnub_ntf_callback_posix.c
```diff
@@ -181,7 +203,7 @@ void* socket_watcher_thread(void *arg)
         }
         pthread_mutex_unlock(&m_watcher.queue_lock);
 
-        monotonic_clock_get_time(&timspec);
+        clock_get_realtime(&timspec);
```
vveljko: The same `timspec` is used a little below, for timers - see `PUBNUB_TIMERS_API`. For that purpose, we _need_ a monotonic clock (a "tick counter", if you will). So, this change (alone) breaks that code. We basically need another variable to hold the monotonic time (that we _also_ need to get).

MacGritsch: Hi @vveljko, below I see that the values are used in `elapsed_ms()` function.
Why should that not work with REALTIME? Do yo mean when system time is changed (twice a year for DST and a few times for leap-second). Same problem exists with `pthread_cond_timedwait` itself.

vveljko: @MacGritsch Yes, as you described, "REALTIME" is actually "wall clock" time. It can change at arbitrary times, going back and forth. So, we can't use that for timers, which measure only elapsed time, not caring about the clock (wall or wrist one :) ).

You are correct, using "REALTIME" for `pthread_cond_timedwait()` is a design flaw in them functions. As you described earlier, POSIX is not clear on this, so that is probably why we assumed that we should use `MONOTONIC` time for it, too. But, we can't change pthreads, so, we must make do with what we have.

MacGritsch: Hi @vveljko,
What do you think about this approach: https://github.com/MacGritsch/c-core/commit/949be7200115dc030bc93639282a0d5f72d2e788?ts=4

vveljko:
This is great, thank you very much.

PR #49, opened 2018-08-02
Update DNS IP from Application

jbansal1510:
This PR is used to enable pubnub library to provide the method to the applications for updating System DNS IP address from the application using pubnub::context.

### DIFF
--- a/cpp/pubnub_common.hpp
+++ b/cpp/pubnub_common.hpp
@@ -16,6 +16,7 @@ extern "C" {
 #include "core/pubnub_timers.h"
 #include "core/pubnub_helper.h"
 #include "core/pubnub_free_with_timeout.h"
+#include "core/pubnub_dns_servers.h"
 #if PUBNUB_PROXY_API
 #include "core/pubnub_proxy.h"
 #endif
@@ -35,7 +36,7 @@ extern "C" {
 #endif
 
 #include <iostream>
-
+#include <sstream>
 
 namespace pubnub {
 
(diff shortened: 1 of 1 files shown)


cpp/pubnub_common.hpp
```diff
@@ -409,6 +410,28 @@ class context {
         set_uuid(pubnub_uuid_to_string(&uuid).uuid);
         return 0;
     }
+    // get the system default dns
```
vveljko: The `pubnub::context` doesn't need these member functions, they don't use the `pubnub_t` which it "hides". So, in the interest of simplicity, they should not be here.

Also, it's not clear that the "default" system DNS is the first DNS you get in the "list" that the `pubnub_dns_read_system_servers_ipv4()` returns, so the name is somewhat misleading.

Instead, you can make a free function `std::vector<std::string> pubnub::read_system_dns_servers()` for ease of use that C++ provides. But setting and getting the primary (and secondary) DNS is something that basically doesn't need a C++ wrapper, as it can't provide any improvements in the interface.

jbansal1510: Hi,
Thanks for the feedback.
I changed the function name to read_system_dns_servers.
As I am not able to access this pubnub_dns_set_primary_server_ipv4_str() function in my application, that's why I have wrote this wrapper function.

Can you explain how to access pubnub_dns_set_primary_server_ipv4_str() function in my application.

jbansal1510:
Hi I am able to do work without changing in pubnub code and there is no change required in pubnub library, so closing the PR.

PR #55, opened 2018-10-09
Develop

vveljko:


vveljko:
The Codacy issue is fake - it's a bug in Codacy, it's not even a false positive.

PR #68, opened 2019-02-15
Add IPv6 support

vveljko:


vveljko:
It's test code, but for manual, ad-hoc testing. It's usual to leave such code in `#if 0` in C. If it becomes more "evolved", we might spin it off to its own module.

qsoftdevelopment:
@vveljko If you ask me, I would move it as we are risking "dirty" code. That is my suggestion. I know it does not slow down the project and that it is not a bug, but more an aesthetic thing.

PR #70, opened 2019-03-06
Add set_transaction_timeout and transaction_timeout_get Qt implementations

mrtryhard:
Sometimes it is useful to modify the default timeout duration which is 10s by default.
The headers appeared to offer the functionality, but it was missing.  This PR offers an implementation.

Sample usage: 
```
pubnub_qt r;
r.set_transaction_timeout(5000); // OK; This will return 0 and sets the transaction
int timeout_s = r.transaction_timeout_get(); // Returns 5000;
r.publish(...); 

// ... 
r.set_transaction_timeout(-20); // Not OK; This will return -1 and keep the previous transaction timeout.
```

### DIFF
--- a/qt/pubnub_qt.cpp
+++ b/qt/pubnub_qt.cpp
@@ -590,6 +590,25 @@ void pubnub_qt::set_ssl_options(ssl_opts options)
 }
 
 
+int pubnub_qt::set_transaction_timeout(int duration_ms)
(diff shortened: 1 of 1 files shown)


vveljko:
Thanks for the PR.

Not really sure how this came to be, but, things happen... :)

PR #71, opened 2019-03-10
Adds support for HTTP Pipelining in Pubnub Qt SDK.

mrtryhard:
This PR allows the SDK client to allow HTTP Pipelining[1] on the QNetworkManager. 

By default, QNetworkManager has this option set at `false`[2]. The proposed implementation respects this by defaulting to `false` as well. 

This should also be working just fine with Pubnub's servers[3].

### Refs
1. https://en.wikipedia.org/wiki/HTTP_pipelining
2. `QNetworkRequest::HttpPipeliningAllowedAttribute`, https://doc.qt.io/qt-5/qnetworkrequest.html
3. https://support.pubnub.com/support/discussions/topics/14000006318

### DIFF
--- a/qt/pubnub_qt.cpp
+++ b/qt/pubnub_qt.cpp
@@ -42,6 +42,7 @@ pubnub_qt::pubnub_qt(QString pubkey, QString keysub)
     , d_transaction_timed_out(false)
     , d_transactionTimer(new QTimer(this))
     , d_use_http_keep_alive(true)
+    , d_use_http_pipelining(false)
     , d_mutex(QMutex::Recursive)
 {
     pbcc_init(d_context.data(), d_pubkey.data(), d_keysub.data());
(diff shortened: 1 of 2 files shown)


vveljko:
Thanks for the PR. 

Each context has only one `d_reply`. IOW, there can only be one open/pending HTTP transaction at a time per context and each context has its own Qt Network Access Manager (for ease of use).

So, I believe that this is not enough and to really implement HTTP pipelining, one would need to have support for it in the Pubnub Qt Context class itself (i.e. support more than one outstanding HTTP request).

mrtryhard:
You're right, I missed that (important) part. I'll investigate if it is not too complex to manage... At the moment, it seems each new request resets the current one. If pipelining is added, the complexity is increased because you can't _just_  do that. It sounds suddenly easy to do errors in that. 

Thanks !

vveljko:
Closing this for now. If you wish, reopen or create a new PR when you figure it out.

PR #79, opened 2019-06-13
Introduce minimal transaction timer duration

vveljko:
Also use QTime for timer duration on Qt ("be more Qt-friendly").


vveljko:
> Should we show some type of warning if they hit they go past the minimum limit?

Changed accordingly.

PR #81, opened 2019-07-06
Add doxygen docs

polarweasel:
If we've already got Doxygen comments and a config file, let's update the doxyfile and add instructions to DIY build the docs.


vveljko:
I'm not sure about this (except the update to the config file). This significantly enlarges/slows down the footprint / `git clone` or tarball download. Also, many people don't really care about Qt, so that would be an annoyance. Not to mention that all other code is also "Doxygen-documented" and some config file/ scripts to build their docs would be nice; if we follow this lead, we would then add _all_ of that documentation in the repository, which would then be a _very significant_ burden on the code.

I propose we build the Doxygen docs on Travis and then attach them to the Release on Github. FWIW, we could then also "bring LaTeX back" and have a PDF variant of the docs (I'm old school, I know :) ).

polarweasel:
Sounds good. I’ll update here to remove the doc output. Can you experiment with adding the doc build to Travis?

HTML only for now. Default PDF output from Doxygen is seriously ugly. We’ll also need to consider at least a little branding on the HTML, and a copyright.

polarweasel:
@vveljko can you take another look? I've made some updates.

PR #84, opened 2019-08-23
fix assertion fail after 24 days

hilbix:
```
24 days =	2 073 600 000 milliseconds
int32 max =	2 147 483 647
25 days =	2 160 000 000 milliseconds
```

Hence we need an int64_t to store the clock value

### DIFF
--- a/lib/msstopwatch/msstopwatch.h
+++ b/lib/msstopwatch/msstopwatch.h
@@ -29,8 +29,8 @@
 #include <stdbool.h>
 
 
-typedef struct { int32_t t_ref; } pbmsref_t;
-typedef int32_t pbms_t;
+typedef struct { int64_t t_ref; } pbmsref_t;
+typedef int32_t pbms_t;		/* Warning!  Overflows after 24d 20h 31m	*/
 
 /** Starts a stopwatch. The result has no particular meaning, 
     is only to be used as the parameter of pbms_elapsed().
--- a/posix/msstopwatch_monotonic_clock.c
+++ b/posix/msstopwatch_monotonic_clock.c
@@ -11,10 +11,11 @@
     clock like POSIX clock_gettime(CLOCK_MONOTONIC,...);
 */
 
-static int32_t msclock(void)
+static int64_t msclock(void)
 {
     struct timespec ts;
     if (0 == monotonic_clock_get_time(&ts)) {
+        /* 32 bit overflows after 2,147,483,648 / 1000 / 24 / 3600 = 24,855 days	*/
         return ts.tv_sec * 1000 + ts.tv_nsec / 1000000;
     }
     else {


hilbix:
This is to fix following:

```
$ ./test1 
Pubnub assert failed 'pbms_active(since)', file '../posix/msstopwatch_monotonic_clock.c', line 47
Aborted (core dumped)
```

Caused by:

```
$ uptime
 12:16:40 up 42 days, 18:28,  7 users,  load average: 0.48, 0.52, 0.50
```

which makes

https://github.com/pubnub/c-core/blob/c3b3674b2c1f130f3899360f22843944770f73b2/posix/msstopwatch_monotonic_clock.c#L14-L18

return a negative `int32_t` which then fails the assertion

https://github.com/pubnub/c-core/blob/c3b3674b2c1f130f3899360f22843944770f73b2/posix/msstopwatch_monotonic_clock.c#L45-L47

due to

https://github.com/pubnub/c-core/blob/c3b3674b2c1f130f3899360f22843944770f73b2/posix/msstopwatch_monotonic_clock.c#L39-L41

hilbix:
BTW, perhaps change `pbms_t` to 64 bit as well to be able to calculate durations longer than 24 days:

https://github.com/pubnub/c-core/blob/0f4b6c4bf4e3693d4cd3ec7f4e7cfd4b42a58b9a/lib/msstopwatch/msstopwatch.h#L32-L33

However I kept it as-is to not possibly break code which depends on the delta being only 32 bit.

vveljko:
Hi @hilbix , thanks for the PR and pointing out the problem. We're currently evaluating alternative solutions, as introducing a 64 bit value makes the code less portable/appropriate for embedded systems. If we don't come up with a good alternative, we'll merge your PR.

hilbix:
FYI PR #85 is an alternative 32-bit-only variant suggestion for this fix.  HTH

vveljko:
Actually, after some further thought, compared to the 32-bit fix, this one is _way_ more simpler and faster on 64 bit CPUs. Thus, it would be nice to add some `#ifdef` to use it on 64 bit CPUs to the other PR #85.

No need to do fancy preprocessor guess-work to see "are we on a 64-bit CPU". Just have it as a `pubnub_config.h` parameter and let the user decide which code to use. She knows what CPU she's on - hell, she might even have a CPU that, even though it's not a 64-bit CPU, this code using `int64_t` works faster!

hilbix:
See my post at #85.  I focus for a 32 bit variant fix (but I cannot promise to do this in-time).

The question here is, after #85 (32 bit variant) is supplied how do we deal with the choice.

- Does the default stay at 32 bit?
- Does the default go for 64 bit?
- Does we have no default, so comile fails if no choice is done?

How to deal with CI?  Does it only compile/test the default variant?  Or does it need to compile+test both.

With the CI thing I probably cannot help at all.  And with the proper decision thing likewise not, too.

vveljko:
@hilbix As mentioned in #85 , we are _very_ grateful to you for bringing this issue to our attention. We'll close these two PRs and work on a fix. Of course, if you come up with a fix before we do, make a new PR, we'll gladly merge it. 😃

PR #85, opened 2019-08-27
fix assertion fail after 24 days (32 bit variant)

hilbix:
```
24 days =	2 073 600 000 milliseconds
int32 max =	2 147 483 647
25 days =	2 160 000 000 milliseconds
```

This variant fixes the problem of the overrun after 24 days.
It does so by not giving back the "uninitialized value" (0) on accident.

On edge cases it makes measurements wrong by +-1ms.
Also there is a rollover after 24 days.  With 32 bit this cannot be circumvented.

Why 0 as uninitialized value?  Because this copes well with "static" initialization.

Why -1 and not 1 as the alternate value when accidentally hitting 0?
Because -1 is a more descriptive "I am on the edge" than just 1.

### DIFF
--- a/posix/msstopwatch_monotonic_clock.c
+++ b/posix/msstopwatch_monotonic_clock.c
@@ -15,10 +15,12 @@ static int32_t msclock(void)
 {
     struct timespec ts;
     if (0 == monotonic_clock_get_time(&ts)) {
-        return ts.tv_sec * 1000 + ts.tv_nsec / 1000000;
+        /* 32 bit overflows after 2,147,483,648 / 1000 / 24 / 3600 = 24,855 days	*/
+        int32_t ret = ts.tv_sec * 1000 + ts.tv_nsec / 1000000;
+        return !ret ? -1 : ret;
     }
     else {
-        return -1;
+        return 0;
     }
 }
 
@@ -32,13 +34,13 @@ pbmsref_t pbms_start(void)
 
 void pbms_stop(pbmsref_t* psw)
 {
-    psw->t_ref = -1;
+    psw->t_ref = 0;
 }
 
 
 bool pbms_active(pbmsref_t stopwatch)
 {
-    return stopwatch.t_ref > -1;
+    return stopwatch.t_ref != 0;
 }
 
 


hilbix:
See #84 for the 64-bit solution.

posix/msstopwatch_monotonic_clock.c
```diff
@@ -15,10 +15,12 @@ static int32_t msclock(void)
 {
     struct timespec ts;
     if (0 == monotonic_clock_get_time(&ts)) {
-        return ts.tv_sec * 1000 + ts.tv_nsec / 1000000;
+        /* 32 bit overflows after 2,147,483,648 / 1000 / 24 / 3600 = 24,855 days	*/
+        int32_t ret = ts.tv_sec * 1000 + ts.tv_nsec / 1000000;
```
vveljko: This is trying to "fix" undefined behavior. Integer overflow is undefined behavior and, while in most cases it does overflow, there's no guarantee what it will do, especially in portable code. 

So, we need to "catch" this before it happens. Something like this (this is both untested and just for illustration, probably wrong):
```
#define MAX_TV_SEC (INT32_MAX / 1000)
#define MAX_TV_NSEC_MSEC_ON_MAX_SEC (INT32_MAX - (MAX_TV_SEC*1000)) 
/*      ^^ a terrible name, I know :) */
...
   if (ts.tv_sec > MAX_TV_SEC) return -1;
   if (ts.tv_sec == MAX_TV_SEC && 
        (ts.tv_nsec/1000000) > MAX_TV_NSEC_MSEC_ON_MAX_TV_SEC) return -1;
   return ts_tv_sec*1000 + ts.tv_nsec / 1000000;
```

hilbix: Good idea, but I then would suggest to do the ceiling at INT_MAX instead.  This way we do not need to change anything in other routines.

vveljko: Did you mean something like:
```
    if ((ts.tv_sec > MAX_TV_SEC) ||
        ((ts.tv_sec == MAX_TV_SEC) && 
        (ts.tv_nsec/1000000) > MAX_TV_NSEC_MSEC_ON_MAX_TV_SEC)) return INT32_MAX;
```
Well, wouldn't that have the effect of time "coming to a still" at `INT32_MAX`? 

Or maybe you meant to "wrap" at INT32_MAX? In that case, this can be made a little simpler by wrapping at `MAX_TV_SEC`, like:
```
    return ((ts.tv_sec % MAX_TV_SEC) * 1000 + ts.tv_nsec / 1000000;
```

Which seems like a nice solution, even if it is a little slow. If needed, it can be optimized to find MAX_TV_NSEC that is a power of two. Also, in this case, a little extra logic needs to be added to `pbms_elapsed`, like:
```
    int32_t elapsed = msclock() - since.t_ref;
    return elapsed + (elapsed < 0) * (MAX_TV_SEC * 1000);
```
Of course this code assumes that it will not actually have been elapsed more than `2*MAX_TV_SEC` since `since.t_ref`, which is probably a rather safe assumption, even though a "paranoid programmer" might add a check for that, too.

hilbix: Yes this is the problem:  `pbms_elapsed` must not return 0 after 24+ days.  But this was what I understood at your code when returning `-1`.  So better forget about that ;)

What I understand now:

- There still is need for a code variant which must be fully 32 bit optimized
- It should be as efficient as possible for embedded platforms
- It must still allow deltas to be calculated, as long as they fit into 31 bit which is about 24 days.  This does not degrade from today where things break.

I do not think we can get to full 32 bit deltas with efficient code.

Also there can be a 64 bit variant, see #84.  But I'd say first focus on the 32 bit variant to get rid of the problem, and then, perhaps, add a more efficient 64 bit variant by default with a choice to go back to 32 bit for selected platforms on implementor's decicion.

Hence this PR here focuses on the 32 bit fix, then afterwards the other PR can be done to do the more efficient implementation on 64 bit.

Doing it step by step probably faster leads to a solution.

The question now is what to optimize more. `msclock` is called most often.  However `pbms_elapsed` probably is called nearly as often, as this is the only operation we have currently on this data type.

So we should expect that `pbms_start`  usually is followed by one or more `pbms_elapsed` where both call `msclock`.

Sorry, but I currently have no good idea for this and do not know if I find the time to do something in this direction.

vveljko: @hilbix I understand. This is indeed a tricky issue. We are _very_ grateful to you for bringing this to our attention and for your efforts in fixing it.

We'll close these two PRs and come up with a fix in a future release.

Hope your experience with C-core brings no more issues in the future. 😄

PR #90, opened 2019-09-23
Fetch history not accepting 0 or 1 as Boolean values

MarcGeh:
Hey guys, 
I had issues using pubnub_history_ex. Setting the arguments 'stringtoken' and 'reverse' to true had no effect. After investigating I found out that apparently the [PubNub REST API for fetching the history](https://www.pubnub.com/docs/pubnub-rest-api-documentation#history) wants to have only true and false as Boolean values, in the code those values were encoded as 1 and 0 however. Changing the macros for those two arguments from APPEND_URL_PARAM_TRIBOOL_M to APPEND_URL_PARAM_BOOL_M fixed the issue.


MarcGeh:
Ok so apparently a unit test failed, since the value is now no longer a tribool (left out when not set). Any suggestions how to fix that?

vveljko:
Thank you reporting this @MarcGeh .

PR #92, opened 2019-10-02
DNS retries, futres fixes

vveljko:
1. If DNS server does not respond, retry. Number of retries is
   a `pubnub_config.h` constant. After that threshold is reached,
   fail the transaction.
2. If a transaction would finish before the user starts to await
   its finish via futres, that await would wait "forever".
3. There is preliminary support for Actions API, but it's not
   fully tested yet, so it's not "officially released".


core/pbcc_actions_api.h
```diff
@@ -0,0 +1,109 @@
...
+#if !defined INC_PBCC_ACTIONS_API
+#define INC_PBCC_ACTIONS_API
+
+#include "pubnub_api_types.h"
+#include "pubnub_memory_block.h"
+
+struct pbcc_context;
+
+
+/** @file pbcc_actions_api.h
```
davidnub: Lol, looks like Message Actions in this. I suppose it's fine, but you can remove it if you're not done. I'm ok either way.

PR #97, opened 2019-11-17
Auto heartbeat posix

vveljko:
User can now enable sending of heartbeat messages "in the background"
on the given context. These heartbeat messages will actually be sent
via a different context, so, won't prevent the user from sending using
her context anyway she sees fit. Heartbeat is sent according to the
latest subscribe() - if the user does a new subscribe() with new
channel/group, the heartbeats will be updated and sent accordingly.

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -1,8 +1,13 @@
 name: c-core
 schema: 1
-version: 2.11.1
+version: 2.12.0
 scm: github.com/pubnub/c-core
 changelog:
+  - version: v2.12.0
+    date: Nov 15, 2019
+    changes:
+      - type: enhancement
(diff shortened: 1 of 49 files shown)


vveljko:
The "errors" reported by Codacy are weird issues that CppCheck has with multiple `#ifdef` configurations. They are hard to fix (teach CppCheck about your precompiler "configurations") and one simply learns to ignore them by default.

davidnub:
> The "errors" reported by Codacy are weird issues that CppCheck has with multiple `#ifdef` configurations. They are hard to fix (teach CppCheck about your precompiler "configurations") and one simply learns to ignore them by default.

Can we just ignore them via compiler flag in Codacy?

PR #104, opened 2020-04-02
Use dynamic allocation with for 'uuid'

parfeon:
## 🌟: use dynamic allocation with for 'uuid'
Dynamically allocate memory for `uuid` in `pubnub_set_uuid` function.


core/pubnub_pubsubapi.h
```diff
@@ -46,7 +46,7 @@ pubnub_t* pubnub_init(pubnub_t* p, const char* publish_key, const char* subscrib
     pubnub_done() on @p p, or the otherwise stop using it (like when
     the whole software/ firmware stops working). So, the contents of
     the @p uuid string is not copied to the Pubnub context @p p.  */
-void pubnub_set_uuid(pubnub_t* p, const char* uuid);
+enum pubnub_res pubnub_set_uuid(pubnub_t* p, const char* uuid);
```
pnehrer, line 49: The comment above suggests that we do not copy the uuid string into the context. But with this change, we now do.

core/pubnub_ccore_pubsub.c
```diff
@@ -115,22 +120,40 @@ char const* pbcc_get_channel(struct pbcc_context* pb)
...
-        PUBNUB_ASSERT_OPT(strlen(uuid) < sizeof pb->uuid);
-        strncpy(pb->uuid, uuid, sizeof pb->uuid);
-        pb->uuid[(sizeof pb->uuid) - 1] = '\0';
+    if (pb->uuid_len > 0) {
+        pb->uuid_len = 0;
+        free(pb->uuid);
     }
-    else {
-        pb->uuid[0] = '\0';
+    pb->uuid = (char *)"";
```
pnehrer, line 129: Why not:
```suggestion
    pb->uuid = NULL;
```

core/pubnub_ccore_pubsub.c
```diff
@@ -115,22 +120,40 @@ char const* pbcc_get_channel(struct pbcc_context* pb)
...
+    if (uuid != NULL) {
+        int len = strlen(uuid);
```
pnehrer, line 131: Should we skip allocation if `len == 0`?

core/pubnub_pubsubapi.h
parfeon: @pnehrer Interesting comment, because it goes against what is inside of function. It was copying `uuid` string even in previous versions: https://github.com/pubnub/c-core/blob/master/core/pubnub_ccore_pubsub.c#L122

```c
void pbcc_set_uuid(struct pbcc_context* pb, const char* uuid)
{
    if (uuid != NULL) {
        PUBNUB_ASSERT_OPT(strlen(uuid) < sizeof pb->uuid);
        strncpy(pb->uuid, uuid, sizeof pb->uuid);
        pb->uuid[(sizeof pb->uuid) - 1] = '\0';
    }
    else {
        pb->uuid[0] = '\0';
    }
}
```

core/pubnub_ccore_pubsub.c
parfeon: @pnehrer I've tried to make as less changes as possible. Fixed-size array has been used before and there was check on first character to be _null character_.
I've searched through code, looks like it is safe to change assignment to `NULL`.

core/pubnub_ccore_pubsub.c
parfeon: @pnehrer woooow, I missed this one. Thanks for noticing this one.

parfeon:
@pnehrer initially context was docs mismatch, where it says, that `uuid` can be 64 chars long, but  code has assertion for 32.
So there was discussion about the way to set UUID and won't lead to memory _wasting_. If we just increase size, this won't be good for platforms, which may be happy even with 8 bytes long `uuid` and has limited amount of memory.
We took approach of dynamic allocation and copy to it. Since it is suggested to use pair of `pubnub_alloc()` and `pubnub_free()`, we can proper resources deallocation.

As for `uuid_len` - if I see something, what modified in rare cases, but used a pretty frequent, I'm trying to enhance stuff related to it. So far, there was [`pb_strnlen_s`](https://github.com/pubnub/c-core/blob/master/lib/pb_strnlen_s.c#L4) which iterate through chars  to find _null char_ and return string length and called twice per action API call. There can be more places, where we would like to calculate `uuid` length. If you think it was premature optimization, I will remove it and stick with `pb_strnlen_s`.

parfeon:
@pnehrer my bad about `pb_strnlen_s`, because it is not only require pointer, but length within which _null char_ search should be performed. Sure, we can specify there 64 chars (before, for `uuid` it used sizeof which during compile-time became 32, because `uuid` was fixed-size).

core/pubnub_ccore_pubsub.c
parfeon: Suggested changes has been added in 644daea.

core/pubnub_ccore_pubsub.c
parfeon: Suggested changes has been added in 644daea.

PR #105, opened 2020-04-07
PubNub SDK 2.12.5 Release

parfeon:
## 🌟: use dynamic allocation for 'uuid'

Dynamically allocate memory for `uuid` in `pubnub_set_uuid` function.

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -1,8 +1,13 @@
 name: c-core
 schema: 1
-version: 2.12.4
+version: "2.12.5"
 scm: github.com/pubnub/c-core
 changelog:
+  - version: v2.12.5
+    date: Apr 8, 2020
+    changes:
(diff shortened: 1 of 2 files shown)


.pubnub.yml
```diff
@@ -1,8 +1,13 @@
 name: c-core
 schema: 1
-version: 2.12.4
+version: "2.12.5"
```
davidnub, line 3: Why quotes?

parfeon: @davidnub matter of habit, since this is how string version (per specs it should be string) appear in some other our SDKs. If important, I can remove double-quotes.

davidnub: Oh. I just noticed that it wasn't there before and wanted to make sure it didn't cause problems processing. If it works then I have no problems with it.

PR #117, opened 2022-01-04
build: fix versions bump file

parfeon:
Fix `versions.json` file format, which cause issues during release.


budgetpreneur (approved):
LGTM 👍

PR #120, opened 2022-03-03
Switch to blocking resolver

parasyte:
- This is the easiest way to allow using the system DNS resolver.
- If Windows has resolv.h, we might be able to use that with the async
  resolver to get the system name servers (usually configured by DHCP,
  but sometimes manually configured).
- We could also look into Windows-specific (winsock2?) APIs for querying
  name servers.

### DIFF
--- a/lib/sockets/pbpal_resolv_and_connect_sockets.c
+++ b/lib/sockets/pbpal_resolv_and_connect_sockets.c
@@ -411,8 +411,6 @@ enum pbpal_resolv_n_connect_result pbpal_resolv_and_connect(pubnub_t* pb)
     hint.ai_canonname                                  = NULL;
     hint.ai_next                                       = NULL;
 
-    prepare_port_and_hostname(pb, &port, &origin);
-
 #ifdef PUBNUB_CALLBACK_API
 #if PUBNUB_PROXY_API
     if (0 != pb->proxy_ipv4_address.ipv4[0]) {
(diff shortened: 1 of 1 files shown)


lib/sockets/pbpal_resolv_and_connect_sockets.c
```diff
@@ -440,48 +450,7 @@ enum pbpal_resolv_n_connect_result pbpal_resolv_and_connect(pubnub_t* pb)
...
-
-    hint.ai_socktype = SOCK_STREAM;
-    hint.ai_family   = AF_UNSPEC;
-    hint.ai_protocol = hint.ai_flags = hint.ai_addrlen = 0;
-    hint.ai_addr                                       = NULL;
-    hint.ai_canonname                                  = NULL;
-    hint.ai_next                                       = NULL;
+#endif /* PUBNUB_CALLBACK_API */
 
     prepare_port_and_hostname(pb, &port, &origin);
```
budgetpreneur, line 455: We have `prepare_port_and_hostname(pb, &port, &origin);` in line#455 and #414. Basically same line of code before and after `#ifdef PUBNUB_CALLBACK_API`. Do we need to call twice?

budgetpreneur: I think, the one at line#414 can be removed.

budgetpreneur:
> * This is the easiest way to allow using the system DNS resolver.
> * If Windows has resolv.h, we might be able to use that with the async
>   resolver to get the system name servers (usually configured by DHCP,
>   but sometimes manually configured).
> * We could also look into Windows-specific (winsock2?) APIs for querying
>   name servers.

For windows, SDK is using `GetNetworkParams` to get DNS servers of host. (`pubnub_dns_system_servers.c -> pubnub_dns_read_system_servers_ipv4()`)
Ref: https://stackoverflow.com/questions/2916675/programmatically-obtain-dns-servers-of-host
resolv.h is mostly for linux based environment.

parasyte:
Closed in favor of #122

PR #122, opened 2022-03-03
Support system name servers in async DNS client

parasyte:
This is an alternative to #120.

The name server is chosen with the following precedence:

1. The server specified with `pubnub_dns_set_primary_server_ipv4()` (assumes `PUBNUB_SET_DNS_SERVERS` is set, which is the default)
2. The server specified with `pubnub_dns_set_secondary_server_ipv4()` (assumes `PUBNUB_SET_DNS_SERVERS` is set, which is the default)
3. The server configured by the system (e.g. DHCP or manually)
4. The hardcoded default server (usually `8.8.8.8`)

If any step fails to find a server, it will try the next in the precedence list. Step 4 always succeeds.

When `PUBNUB_SET_DNS_SERVERS` is _not_ set, the name server configured on the system is always chosen first.

This patch also fixes a minor memory leak in the Windows-specific name server lookup function when an error occurs.

### DIFF
--- a/lib/sockets/pbpal_resolv_and_connect_sockets.c
+++ b/lib/sockets/pbpal_resolv_and_connect_sockets.c
@@ -84,6 +84,13 @@ static void prepare_port_and_hostname(pubnub_t*    pb,
 
 
 #ifdef PUBNUB_CALLBACK_API
+
+static void get_default_dns_ip(struct pubnub_ipv4_address *addr) {
+    if (pubnub_dns_read_system_servers_ipv4(addr, 1) != 1) {
+        inet_pton(AF_INET, PUBNUB_DEFAULT_DNS_SERVER, addr);
+    }
+}
+
 #if PUBNUB_SET_DNS_SERVERS
 #if PUBNUB_CHANGE_DNS_SERVERS
 static void get_dns_ip(struct pbdns_servers_check* dns_check, struct sockaddr* addr)
@@ -104,7 +111,7 @@ static void get_dns_ip(struct pbdns_servers_check* dns_check, struct sockaddr* a
 #if PUBNUB_USE_IPV6
             addr->sa_family = AF_INET6;
 #else
-            inet_pton(AF_INET, PUBNUB_DEFAULT_DNS_SERVER, p);
+            get_default_dns_ip((struct pubnub_ipv4_address*)p);
 #endif /* PUBNUB_USE_IPV6 */
         }
     }
@@ -119,7 +126,7 @@ static void get_dns_ip(struct pbdns_servers_check* dns_check, struct sockaddr* a
                 || (dns_check->dns_server_check & dns_check->dns_mask)) {
                 dns_check->dns_mask <<= 1;
                 addr->sa_family = AF_INET;
-                inet_pton(AF_INET, PUBNUB_DEFAULT_DNS_SERVER, p);
+                get_default_dns_ip((struct pubnub_ipv4_address*)p);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


lib/sockets/pbpal_resolv_and_connect_sockets.c
```diff
@@ -119,7 +126,7 @@ static void get_dns_ip(struct pbdns_servers_check* dns_check, struct sockaddr* a
                 || (dns_check->dns_server_check & dns_check->dns_mask)) {
                 dns_check->dns_mask <<= 1;
                 addr->sa_family = AF_INET;
-                inet_pton(AF_INET, PUBNUB_DEFAULT_DNS_SERVER, p);
+                get_default_dns_ip((struct pubnub_ipv4_address*)p);
```
budgetpreneur, line 129: This is under the condition `#if PUBNUB_USE_IPV6` . Should it be `pubnub_ipv6_address` instead of `pubnub_ipv4_address`?

lib/sockets/pbpal_resolv_and_connect_sockets.c
```diff
@@ -148,7 +155,7 @@ static void get_dns_ip(struct sockaddr* addr)
             && (pubnub_get_dns_secondary_server_ipv6((struct pubnub_ipv6_address*)pv6)
                 == -1)) {
             addr->sa_family = AF_INET;
-            inet_pton(AF_INET, PUBNUB_DEFAULT_DNS_SERVER, p);
+            get_default_dns_ip((struct pubnub_ipv4_address*)p);
```
budgetpreneur, line 158: if statement is checking with `(struct pubnub_ipv6_address*)pv6` but inside that code is using` (struct pubnub_ipv4_address*)p` . Can you please clarify, if that is ok.

budgetpreneur: Shouldn't that be `get_default_dns_ip((struct pubnub_ipv6_address*)pv6);` ?

lib/sockets/pbpal_resolv_and_connect_sockets.c
```diff
@@ -157,9 +164,8 @@ static void get_dns_ip(struct sockaddr* addr)
 static void get_dns_ip(struct sockaddr* addr)
 {
     addr->sa_family = AF_INET;
-    inet_pton(AF_INET,
-              PUBNUB_DEFAULT_DNS_SERVER,
-              &(((struct sockaddr_in*)addr)->sin_addr.s_addr));
+    struct pubnub_ipv4_address* p = &(((struct sockaddr_in*)addr)->sin_addr.s_addr));
```
budgetpreneur, line 167: It seems code is using the same function for both IPV4 and IPv6.  Please confirm if this works for IPv6?

lib/sockets/pbpal_resolv_and_connect_sockets.c
parasyte: No, I don't think so. See line 128 above:

```c
addr->sa_family = AF_INET;
```

This sets the address family to IPv4. For another comparison, line 112 sets it to IPv6:

```c
addr->sa_family = AF_INET6;
```

lib/sockets/pbpal_resolv_and_connect_sockets.c
parasyte: Same answer here as on line 129. The code expects IPv4 here, due to setting the address family.

lib/sockets/pbpal_resolv_and_connect_sockets.c
parasyte: Yes, it will be fine for IPv6, as long as `8.8.8.8` was always working for IPv6 before this patch! If there were preexisting bugs with that setup, we can address them. I don't think I enabled IPv6 in my testing, but I have full IPv6 support on my network. I can make some additional passes to ensure it's all good.

parasyte: I just learned that IPv6 is already enabled by default in the makefiles. I am satisfied that it works as intended. Would you agree?

PR #124, opened 2022-03-04
Consolidation of DNS resolve and memory related fixes

budgetpreneur:
fix: Support system name servers in async DNS client

Support system name servers in async DNS client

fix: Fix multiple memory safety and leak issues

Fix multiple memory safety and leak issues

fix: Fix slash char encoding for pnsdk

Fix slash char encoding for pnsdk

### DIFF
--- a/core/pubnub_crypto.c
+++ b/core/pubnub_crypto.c
@@ -115,7 +115,7 @@ int pubnub_encrypt(char const* cipher_key, pubnub_bymebl_t msg, char* base64_str
     #endif
 
     int max_size = base64_max_size(encrypted.size);
-    if (*n + 1 < max_size) {
+    if (*n + 1 < (size_t)max_size) {
         PUBNUB_LOG_DEBUG("base64encode needs %d bytes but only %zu bytes are available\n", max_size, *n);
         return -1;
     }
@@ -166,7 +166,7 @@ int pubnub_encrypt_buffered(char const* cipher_key, pubnub_bymebl_t msg, char* b
     #endif
 
     int max_size = base64_max_size(buffer.size);
-    if (*n + 1 < max_size) {
+    if (*n + 1 < (size_t)max_size) {
         PUBNUB_LOG_DEBUG("base64encode needs %d bytes but only %zu bytes are available\n", max_size, *n);
         return -1;
     }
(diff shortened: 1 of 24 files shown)


core/pubnub_crypto.c
```diff
@@ -115,7 +115,7 @@ int pubnub_encrypt(char const* cipher_key, pubnub_bymebl_t msg, char* base64_str
     #endif
 
     int max_size = base64_max_size(encrypted.size);
-    if (*n + 1 < max_size) {
+    if (*n + 1 < (size_t)max_size) {
```
parasyte, line 118: This is ok, but maybe the `base64_max_size` function definition should change to return `size_t` to avoid these casts. I feel like `size_t` is a more appropriate type for allocation sizes than `int`.

The only reason I wrote it to return `int` is because it accepts an `int` input, matching the `base64encode` definition, whose two size inputs could also be changed to `size_t`. The `int` return value on `base64encode` is for error handling.

https://github.com/pubnub/c-core/blob/4a5896634b91ef7ca4ec8676d2d090d664173556/core/pubnub_crypto.h#L145-L153

core/pubnub_grant_token_api.c
```diff
@@ -99,40 +99,6 @@ pubnub_chamebl_t pubnub_get_grant_token(pubnub_t* pb)
...
-    length = decoded.size;
-
-    CborParser parser;
-    CborValue it;
-    
-    char * json_result = (char*)malloc(5*(strlen(rawToken)/4));
-    sprintf(json_result, "%s", "");
-    CborError err = cbor_parser_init(buf, length, 0, &parser, &it);
-    if (!err){
-        data_recursion(&it, 1, json_result);
```
parasyte, line 128: Other way to fix this without moving a large chunk of code (making diff harder to review) is adding a function definition for `data_recursion` on line 101 or higher. Usually, the function definitions appear in header files, but then you may have scoping issues (like accidentally exposing internal functions and types to external callers) and trigger other errors for other language quirks like not being allowed to define a function more than once.

There are workarounds for these issues, as well, but if `data_recursion` is only referenced in this file, it's fine to define it in this file.

parasyte (approved):

kleewho (approved):

budgetpreneur:

PR #129, opened 2022-06-29
removed extra parenthesis and applied struct cast

budgetpreneur:
fix: Removed extra parenthesis in get_dns_ip function code

Removed extra parenthesis in get_dns_ip function code


budgetpreneur:
Taken care in PR#130. So closing this.

PR #130, opened 2022-07-01
Missing UUID query param

budgetpreneur:
fix: Removed extra parenthesis in get_dns_ip function code

Removed extra parenthesis in get_dns_ip function code

fix: Added uuid query param to history, set/get state, wherenow, channel-group operations

Added uuid query param to history, set/get state, wherenow, channel-group operations

### DIFF
--- a/core/pubnub_ccore.c
+++ b/core/pubnub_ccore.c
@@ -218,6 +218,7 @@ enum pubnub_res pbcc_history_prep(struct pbcc_context* pb,
                                   char const*          end)
 {
     char const* const uname = pubnub_uname();
+    char const* uuid = pbcc_uuid_get(pb);
     enum pubnub_res rslt = PNR_OK;
 
     pb->http_content_len = 0;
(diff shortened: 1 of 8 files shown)


core/pubnub_ccore.c
```diff
@@ -385,6 +388,7 @@ enum pubnub_res pbcc_where_now_prep(struct pbcc_context* pb, const char* uuid)
                                 uuid);
     URL_PARAMS_INIT(qparam, PUBNUB_MAX_URL_PARAMS);
     if (pubnub_uname()) { ADD_URL_PARAM(qparam, pnsdk, pubnub_uname()); }
+    if (pb_uuid) { ADD_URL_PARAM(qparam, uuid, pb_uuid); }
```
parfeon, line 391: @budgetpreneur can we have here and in other places this line:
```
char const* const uname = pubnub_uname();
if (uname) { ADD_URL_PARAM(qparam, pnsdk, uname); }
```
where, `pnsdk` should be set instead of:
```
if (pubnub_uname()) { ADD_URL_PARAM(qparam, pnsdk, pubnub_uname()); }
```
?

I know, it is not related to scope of this change, just maybe :)

parfeon (approved):
LGTM!

core/pubnub_ccore.c
budgetpreneur: ok..made changes per your comment.

parfeon: Thanks!

budgetpreneur:

PR #136, opened 2022-10-07
Fix token memory leak

Xavrax:
fix: fix memory leak in cpp `parse_token` method

fix: fix buffer overflow in core `pubnub_parse_token` function for some cases

fix: fix buffer overflow in core `pubnub_encrypt` function for randomized initial vector

### DIFF
--- a/core/pubnub_grant_token_api.c
+++ b/core/pubnub_grant_token_api.c
@@ -18,6 +18,7 @@
 #include "core/pubnub_crypto.h"
 
 #include <ctype.h>
+#include <stdlib.h>
 #include <string.h>
 #include "lib/base64/pbbase64.h"
 #ifdef _MSC_VER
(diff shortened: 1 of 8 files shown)


core/pubnub_grant_token_api.c
```diff
@@ -99,13 +100,32 @@ pubnub_chamebl_t pubnub_get_grant_token(pubnub_t* pb)
     return result;
 }
 
+static unsigned int safe_alloc_strcat(char* destination, const char* source, unsigned int current_allocation_size) {
+    unsigned int concat_size = strlen(destination) + strlen(source);
+    if (concat_size > current_allocation_size) {
+	unsigned int new_allocation_size = (5 * current_allocation_size) / 4; // +20%
+	char* new_alloc = (char*)malloc(new_allocation_size);
```
budgetpreneur, line 107: just checking if we need to free "new_alloc" or not because we did malloc?

Xavrax: We do not need to free `new_alloc` because it just replacing another allocation. `destination` allocation is freed somewhere else so only what we have to do is to clean old allocation in case of replacement.

budgetpreneur (approved):
LGTM. Tested on windows with the code changes and it is working fine.

mateusz-dahlke:

PR #146, opened 2022-12-01
fix MSVC compiler for parse_token

Xavrax:
fix: MSVC compilation error

Fixed compilation error for MSVC in `pubnub_parse_token` function.

### DIFF
--- a/core/pubnub_grant_token_api.c
+++ b/core/pubnub_grant_token_api.c
@@ -201,7 +201,7 @@ static CborError data_recursion(CborValue* it, int nestingLevel, char** json_res
                         encoded_sig.ptr[0] = '\0';
                     }
 
-                    char base64_str[encoded_sig.size + 2];
+                    char base64_str[67]; // HMAC+SHA256 max size + quotes + tailing null
                     sprintf(base64_str, "\"%s\"", encoded_sig.ptr);
 
                     free(encoded_sig.ptr);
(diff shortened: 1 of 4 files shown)


core/pubnub_grant_token_api.c
```diff
@@ -201,7 +201,7 @@ static CborError data_recursion(CborValue* it, int nestingLevel, char** json_res
                         encoded_sig.ptr[0] = '\0';
                     }
 
-                    char base64_str[encoded_sig.size + 2];
+                    char base64_str[67]; // HMAC+SHA256 max size + quotes + tailing null
```
budgetpreneur, line 204: any reason for hardcoding here with 67?

budgetpreneur (approved):
I am able to build this PR for grant_token/parse_token example. 👍 
Commented on hardcoded value.

core/pubnub_grant_token_api.c
Xavrax: According to the [docs](https://www.pubnub.com/docs/sdks/rest-api#grant-access-v3) the `sig` is an HMAC+SHA256 value. Usually we got around 48 bytes but the maximum value of that key can be up to [64 bytes](https://crypto.stackexchange.com/questions/34864/key-size-for-hmac-sha256).

mateusz-dahlke:

pn-op:
🚀 Release successfully completed 🚀

PR #148, opened 2022-12-21
Fix command handler workflow

parfeon:
fix(github-actions): fix command handler workflow

Fix how command handler workflow verify whether comment is command for bot or not.


Xavrax (approved):
Seems good.

mateusz-dahlke (approved):

PR #150, opened 2023-01-06
cpp - nmake build error fix for fetchhistory

budgetpreneur:


Xavrax (approved):
Nice catch! - we should be aware of the lack of the configuration options!

mateusz-dahlke (approved):

PR #151, opened 2023-01-11
add state param to heartbeat

budgetpreneur:


Xavrax:
I think that this params should be optional and hidden by some configuration flag. I'm sure that not all customers needs this change.

budgetpreneur:
requirements changed. PR no longer good.

PR #155, opened 2023-01-16
Create `set_state_ex` stub implementation for QT

Xavrax:
After #154 

Stub implementation because QT is not working at all.


budgetpreneur (approved):
Unable to test the changes since Qt is broken. Approving this PR as placeholder reminder for fixes.

PR #160, opened 2023-05-24
OpenSSL deprecation fix

Xavrax:
refactor(openssl): use newest openssl API
Conditionally use of using newest openssl API

### DIFF
--- a/openssl/pbpal_connect_openssl.c
+++ b/openssl/pbpal_connect_openssl.c
@@ -239,7 +239,11 @@ enum pbpal_tls_result pbpal_check_tls(pubnub_t* pb)
     PUBNUB_LOG_TRACE("pb=%p: SSL connected\n", pb);
     socket_set_rcv_timeout(pb->pal.socket, pb->transaction_timeout_ms);
 
+    #if OPENSSL_VERSION_NUMBER < 0x30000000L
     cert = SSL_get_peer_certificate(ssl);
+    #else
+    cert = SSL_get0_peer_certificate(ssl);
+    #endif
     if (cert != NULL) {
         rslt = SSL_get_verify_result(ssl);
         X509_free(cert);
--- a/openssl/pbpal_openssl.c
+++ b/openssl/pbpal_openssl.c
@@ -17,6 +17,11 @@
 #include <string.h>
 
 #include <openssl/opensslv.h>
+#include <openssl/crypto.h>
+
+// TODO: find good definition
+#include <openssl/ssl.h>
+#include <openssl/crypto.h>
 
 #define HTTP_PORT 80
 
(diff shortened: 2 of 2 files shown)


openssl/pbpal_connect_openssl.c
```diff
@@ -239,7 +239,11 @@ enum pbpal_tls_result pbpal_check_tls(pubnub_t* pb)
     PUBNUB_LOG_TRACE("pb=%p: SSL connected\n", pb);
     socket_set_rcv_timeout(pb->pal.socket, pb->transaction_timeout_ms);
 
+    #if OPENSSL_VERSION_NUMBER < 0x30000000L
     cert = SSL_get_peer_certificate(ssl);
+    #else
+    cert = SSL_get0_peer_certificate(ssl);
```
parfeon, line 245: Why `SSL_get0_peer_certificate` used in place of `SSL_get_peer_certificate`?
According to docs:
> SSL_get_peer_certificate() is an alias of SSL_get1_peer_certificate().

`SSL_get0_peer_certificate` doesn't increase the reference count on X509 object, and this doesn't need to be freed to it. In the current implementation, there is release on line **249** (`X509_free(cert)`).

```suggestion
    #if OPENSSL_VERSION_NUMBER < 0x30000000L
    cert = SSL_get_peer_certificate(ssl);
    #else
    cert = SSL_get1_peer_certificate(ssl);
```

openssl/pbpal_openssl.c
```diff
@@ -111,6 +115,14 @@ static int pal_init(void)
         if (locks_setup()) {
             return -1;
         }
+        #else // OPENSSL_VERSION_NUMBER >= 0x10100000L
+        OPENSSL_init_ssl(OPENSSL_INIT_LOAD_SSL_STRINGS, NULL);
+        OPENSSL_init_crypto(
+                OPENSSL_INIT_ADD_ALL_CIPHERS | OPENSSL_INIT_ADD_ALL_DIGESTS,
+                NULL
+        );
```
parfeon, line 123: If we want to set up this manually, are there any reasons why other default options wasn't included?
`OPENSSL_INIT_LOAD_CRYPTO_STRINGS ` – this one will include error strings, for example.

Docs has more flags marked as default.

parfeon (changes requested):

openssl/pbpal_connect_openssl.c
Xavrax: > SSL_get_peer_certificate() is an alias of SSL_get1_peer_certificate().

I didn't find that info. 
But anyway - calling the `get1` we have to free that function at the end.

Xavrax: we free that anyway - my bad

parfeon (approved):
LGTM!

Xavrax:
@pubnub-release-bot release

PR #168, opened 2023-11-09
fix free on not initialised PubNub

Xavrax:
fix: fix `pubnub_free()` function on not initialised PubNub
Fix `pubnub_free()` function on not initialised PubNub that can cause exceptions/undefined behaviours

### DIFF
--- a/core/pubnub_alloc_std.c
+++ b/core/pubnub_alloc_std.c
@@ -135,11 +135,17 @@ void pballoc_free_at_last(pubnub_t* pb)
 
 int pubnub_free(pubnub_t* pb)
 {
-    int result = -1;
-
     PUBNUB_ASSERT(pb_valid_ctx_ptr(pb));
 
     PUBNUB_LOG_TRACE("pubnub_free(%p)\n", pb);
+
+    if (pb->state == PBS_NULL) {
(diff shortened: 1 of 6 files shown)


core/pubnub_alloc_std.c
```diff
@@ -135,12 +135,18 @@ void pballoc_free_at_last(pubnub_t* pb)
...
-    int result = -1;
-
     PUBNUB_ASSERT(pb_valid_ctx_ptr(pb));
 
     PUBNUB_LOG_TRACE("pubnub_free(%p)\n", pb);
 
+    if (pb->state == PBS_NULL) {
+        PUBNUB_LOG_TRACE("pubnub_free(%p) not initialized - freeing...\n", pb);
+        remove_allocated(pb);
+        free(pb);
```
parfeon, line 145: Wouldn't there be over release with `remove_allocated`?

Xavrax: As i understood -`remove_allocated` only removes concrete instance from the memory. Nothing more.

Xavrax: Also it frees only the whole list of contextes if there is nothing left.

parfeon (approved):
LGTM!

PR #170, opened 2023-11-22
Handle unencrypted message while getting messages with crypto

Xavrax:
fix: Handle unencrypted message while getting messages with crypto

Handle unencrypted message while getting messages with crypto

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -1,8 +1,13 @@
 name: c-core
 schema: 1
-version: "4.7.0"
+version: "4.7.1"
 scm: github.com/pubnub/c-core
 changelog:
+  - date: 2023-11-23
+    version: v4.7.1
+    changes:
+      - type: bug
+        text: "Handle unencrypted message while getting messages with crypto."
   - date: 2023-11-20
     version: v4.7.0
     changes:
@@ -763,7 +768,7 @@ sdks:
             distribution-type: source code
             distribution-repository: GitHub release
             package-name: C-Core
-            location: https://github.com/pubnub/c-core/releases/tag/v4.7.0
+            location: https://github.com/pubnub/c-core/releases/tag/v4.7.1
             requires:
               -
                 name: "miniz"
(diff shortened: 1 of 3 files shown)


core/pubnub_ccore_pubsub.c
```diff
@@ -122,8 +122,8 @@ char const* pbcc_get_msg(struct pbcc_context* pb)
             if (NULL != pb->crypto_module) {
                 char* trimmed = (char*)malloc(strlen(rslt) + 1); // same length as rslt
                 if (NULL == trimmed) {
-                    PUBNUB_LOG_ERROR("pbcc_get_msg(pbcc=%p) - failed to allocate memory for trimmed string. Dropping message!\n", pb);
-                    return NULL;
```
parfeon, line 126: I'm not really sure about this one because the reason of error is not issues with source message. We need to report issues with decryption. Allocation or buffer size should be treated the same.

core/pubnub_ccore_pubsub.c
```diff
@@ -133,20 +133,20 @@ char const* pbcc_get_msg(struct pbcc_context* pb)
...
                 if (NULL == rslt_block.ptr) {
-                    PUBNUB_LOG_ERROR("pbcc_get_msg(pbcc=%p) - decryption failed. Dropping message!\n", pb);
-                    return NULL;
+                    PUBNUB_LOG_WARNING("pbcc_get_msg(pbcc=%p) - decryption failed. Returning original message!\n", pb);
+                    return rslt;
                 }
 
                 if (pb->decrypted_message_count >= PUBNUB_MAX_DECRYPTED_MESSAGES) {
-                    PUBNUB_LOG_ERROR("pbcc_get_msg(pbcc=%p) - maximum number of decrypted messages reached. Dropping message!\n", pb);
-                    return NULL;
```
parfeon, line 149: The reason is the same above. Probably the original implementation is correct one.

parfeon (approved):
Mostly good but unsure about two moments.

core/pubnub_ccore_pubsub.c
Xavrax: Okay, so Should I return just error with NULL, yep?

core/pubnub_ccore_pubsub.c
Xavrax: as above

Xavrax:
@pubnub-release-bot release

PR #171, opened 2023-12-07
provide `#if` macros for pn features to not rely only on makefiles

Xavrax:
feat: Add `#if` switches into files that are related to PubNub features to not rely only on makefiles.

Add `#if` switches into files that are related to PubNub features to not rely only on makefiles. [Be careful when update. It's not a breaking change at all but might fail build for custom makefiles!]


Xavrax:
@pubnub-release-bot release

PR #172, opened 2023-12-22
cmake support

Xavrax:
feat: CMake support

Provide CMake support

feat: adjust `build.cs` for CMake

Adjust `build.cs` unreal engine file for CMake build


tomadamatkinson:
This would be a great addition! :+1:

CMakeLists.txt
```diff
@@ -0,0 +1,376 @@
...
+        ${FEATURE_SOURCEFILES} 
+        ${CMAKE_CURRENT_LIST_DIR}/core/pbcc_revoke_token_api.c 
+        ${CMAKE_CURRENT_LIST_DIR}/core/pubnub_revoke_token_api.c)
+endif()
+
+set(OPENSSL_LINK_LIBRARIES)
+if(${OPENSSL})
+    message(STATUS "Using OpenSSL")
+
+    if(${OPENSSL_ROOT_DIR} STREQUAL "")
```
tomadamatkinson, line 247:
```suggestion
    if("${OPENSSL_ROOT_DIR}" STREQUAL "")
```

This line is failing locally for me. I've seen that its set at the top of the file but locally on my system this isnt initialised in time. This might be due to how CMakes caching works. The above suggestion works as expected

tomadamatkinson: Compiles out of the box on Ubuntu 22.04! Nice work

Xavrax: Thank you for your efforts!

tomadamatkinson: Can confirm it also works out of box using CPM which is a wrapper of CMake ExternalProject without issue and as a submodule

```
CPMAddPackage(
  NAME pubnub
  GIT_REPOSITORY https://github.com/pubnub/c-core
  GIT_TAG feat/cmake-build-system
)
```

tomadamatkinson:
@Xavrax created #173 as there are a few more involved changes for this to work with an external project which i presume is one of the goals of this CMake feature. Symbols and includes are missing

Is there a plan to add CMake support for the cpp module?

Xavrax:
> @Xavrax created #173 as there are a few more involved changes for this to work with an external project which i presume is one of the goals of this CMake feature. Symbols and includes are missing
> 
> Is there a plan to add CMake support for the cpp module?

Yeah. This is very first iteration of CMake support. We want to go further with those changes to move out from raw makefiles. 
Ofc. I can't tell you when the C++ support for CMake will be avaliable but I will point   out the need for it. 
Also this change for cpp is not a big deal (I think) so maybe I will be able to prepare that in some free time. :)

tomadamatkinson:
> Ofc. I can't tell you when the C++ support for CMake will be avaliable but I will point out the need for it.
Also this change for cpp is not a big deal (I think) so maybe I will be able to prepare that in some free time. :)

Sounds good to me. Heres a start #174. Feel free to build on it. I likely wont touch this for a few days

tomadamatkinson:
Interesting. I guess the CI didnt run on my local PR because they didnt point at main :thinking:

parfeon:
@tomadamatkinson for security reasons, CI won't run on forked branches.

lib/pb_extern.h
```diff
@@ -6,6 +6,10 @@
 /// `PUBNUB_EXTERN` before including any Pubnub header.
 
 #ifndef PUBNUB_EXTERN
+#ifdef __cplusplus
+#define PUBNUB_EXTERN extern "C"
```
tomadamatkinson, line 10: As the CI tests are using the Makefile version of the build system it is likely this define change which is breaking the CI

tomadamatkinson: Not sure how it would break. Surely exposing C interfaces should be demangled?

Xavrax: But anyway it is a little but strange as our unit tests are written in C and compiled by gcc so the behaviour shouldn't change. ;o

lib/pb_extern.h
```diff
@@ -6,6 +6,10 @@
...
+#else
 #define PUBNUB_EXTERN extern
+#endif
```
tomadamatkinson: For now its important that the CI passes. This change should fix things. I tested locally with both Make and CMake and using the simple external demo. All worked correctly.

```suggestion
#define PUBNUB_EXTERN extern
```

tomadamatkinson: Oops you get there before i did! Sorry i should have refreshed the page first :laughing:

Xavrax: Bad timing :D

tomadamatkinson: Just noticed that too. Best closing the two open threads and moving onto the next problems! :rofl: 

Sorry for the confusion. Sometimes github is kind and updates threads automatically... sometimes it doesn't

Xavrax: Okay. I removed the extern.
I forgot that we wrap the whole C code base into `extern "C" {}` block. :/

tomadamatkinson:
I was just writing a comment with the same suggestion. Nice work. I think [working!](https://github.com/pubnub/c-core/pull/172/commits/871686444c5ecd5fa84612ebb6d7f85356336dc6) will work!

Xavrax:
Yeah! Seems like CMake is almost ready. 
As I'm finishing for now and I will be off tomorrow - you can expect the release on Monday (I have to check one more thing before but I don't have time to do it now and also @parfeon can find some additional stuff to fix ;d). I will notify you if the release will be delayed. 

@tomadamatkinson Thank you so much for the whole collaboration! <3

parfeon (approved):
LGTM!

Xavrax:
@pubnub-release-bot release

PR #173, opened 2024-01-03
CMake External Project Fixes

tomadamatkinson:
I tested the new CMake branch. The build works great but I ran into some issues when integrating PubNub into an external CMake project

Demo project is set up as shown

image

CMake build fails on missing symbols and missing headers

PR includes fixes for symbols and uses the preferred target based headers `target_include_directories`. To do this i moved the source file list creation above the target `add_library` and moved the header includes below the target creation.

PR #174, opened 2024-01-03
CMake CPP Support

tomadamatkinson:
Builds on top of #173 and so should not merge before 173.

Adds CPP samples and supports sync and callback apis

Adds an options overview so its easier to see which features are configured

image

TODO:

- Support tests
- Fix segfault

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,13 +9,31 @@ message(STATUS Preparing\ ${PROJECT_NAME}\ version\ ${PROJECT_VERSION})
 function(num_option name description default)
     option(${name} ${description} ${default})
     if(${name})
+        message(STATUS "[*] ${name}")
         set(${name} 1 PARENT_SCOPE)
     else()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -9,14 +9,32 @@ message(STATUS Preparing\ ${PROJECT_NAME}\ version\ ${PROJECT_VERSION})
 function(num_option name description default)
     option(${name} ${description} ${default})
     if(${name})
+        message(STATUS "[*] ${name}")
         set(${name} 1 PARENT_SCOPE)
     else()
+        message(STATUS "[ ] ${name}")
```
Xavrax, line 15: Small thing but I love it. Much better than that flag message I've created before!

tomadamatkinson:
@Xavrax if we are merging into the cmake branch then the TODO's on here can be resolved in separate targeted PRs or commits

Xavrax (approved):
Thank you for the work! 
I'm going to finish that asap.

PR #176, opened 2024-01-30
fix building process for Unreal Enigne

Xavrax:
fix: fix readme about Windows build commands
Fix building path for Windows environment

fix: Unreal paths for Windows builds
Fix Unreal paths for Windows builds


parfeon (approved):
LGTM!

PR #179, opened 2024-03-27
provide a way to handle custom openssl paths

Xavrax:
build: provide a way to handle custom openssl paths

provide a way to handle custom openssl paths


.github/workflows/release/versions.json
```diff
@@ -7,6 +7,6 @@
     { "pattern": "#define PUBNUB_SDK_VERSION \"(.+)\"$", "cleared": true }
   ],
   "CMakeLists.txt": [
-    { "pattern": "^version(\"(.+)\")$", "cleared": true }
+    { "pattern": "project(c-core VERSION ((.+))$", "cleared": true }
```
parfeon, line 10: If this is a correct pattern, it will be applied in the release after this one (manual update will be required for this release).

parfeon (approved):
LGTM!

.github/workflows/release/versions.json
Xavrax: Yeah. I've made a mistake some time ago. 
I don't need to release that PR as there are no changes in the code itself.

PR #182, opened 2024-04-16
esp32 support

Xavrax:
feat: ESP32 platform support
Provided support for ESP32 devices via ESP-IDF framework

feat: MBedTLS support
Provided support for MBedTLS library used within esp32 platform

refactor: `strcpy` instead of `strncpy` when it is safe
Replace `strncpy` with `strcpy` in blocks where it is safer to be used.


CMakeLists.txt
```diff
@@ -430,43 +477,83 @@ set(SOURCEFILES
...
-    elseif(WIN32 OR WIN64 OR MSVC)
-        target_include_directories(pubnub PUBLIC
-            ${CMAKE_CURRENT_LIST_DIR}/windows
-            ${CMAKE_CURRENT_LIST_DIR}/core/c99)
+    message(STATUS "Building ESP32 IDF component")
+
+    set(EXTRA_CFLAGS ${EXTRA_CFLAGS} ${FLAGS})
+
+    if (${MBEDTLS})
+       idf_component_register(SRCS ${SOURCEFILES}
```
blake-spangenberg, line 526: When testing this branch with our existing codebase, the following files required by our project were not part of `SOURCEFILES` and thus not being compiled:

```cmake
${CMAKE_CURRENT_LIST_DIR}/mbedtls/pbpal_mbedtls.c
${CMAKE_CURRENT_LIST_DIR}/mbedtls/pbpal_connect_mbedtls.c
${CMAKE_CURRENT_LIST_DIR}/freertos/pbtimespec_elapsed_ms_esp32.c
```

CMakeLists.txt
blake-spangenberg, line 1: For our project to compile properly using `MBEDTLS` and `ESP_PLATFORM`, either:
- Some `-Wall` GCC compilation errors need to be resolved
- Bypassing that compilation requirement by including this line after `idf_component_register`:

```cmake
target_compile_options(${COMPONENT_LIB} PRIVATE -w)
```

blake-spangenberg (changes requested):
Greetings,

I'm working with Insteon and am eagerly awaiting the completion of esp support for the PubNub SDK. Seeing as this port is nearing completion, I have experimented with integrating it into our project have have some feedback as a result of that experiment:

- In `core/pubnub_alloc_static.c`, `pubnub_alloc` relies on local static FreeRTOS mutexes, but there is nothing in that file which calls `pubnub_mutex_init` on them, thus causing an exception at runtime for our project:

```c
pubnub_t *pubnub_alloc(void)
{
    size_t i;

    for (i = 0; i < PUBNUB_CTX_MAX; ++i) {
        // Something like this is needed...
        // if (!m_aCtx[i].monitor) {
        //    pubnub_mutex_init(m_aCtx[i].monitor)
        // }
        pubnub_mutex_lock(m_aCtx[i].monitor);
        if (m_aCtx[i].state == PBS_NULL) {
            m_aCtx[i].state = PBS_IDLE;
            pubnub_mutex_unlock(m_aCtx[i].monitor);
            return m_aCtx + i;
        }
        pubnub_mutex_unlock(m_aCtx[i].monitor);
    }

    return NULL;
}
```

- Would you consider incorporating this into espressif’s [Component Registry](https://components.espressif.com/)? Their guide is provided [here](https://docs.espressif.com/projects/idf-component-manager/en/latest/guides/packaging_components.html). This would be a great opportunity for you as it could make your service much more discoverable to those exploring the ESP32 for future projects. For us, it would greatly simplify the process of keeping up-to-date with the latest release of your SDK.

Xavrax:
Hello @blake-spangenberg !
At the end of the week I've changed a little be a source to test the SDK over the qemu emulator.
I was able to test our pub/sub utility without any errors. 
Right now I'm polishing the code to ensure that it works with other features and remove all the magic hacks and totos I've made during the process. 

We're planning to release it this week but if you will notice any other issues with the code - let us know about it. We want to be sure that our product satisfies our customer needs <3.

CMakeLists.txt
Xavrax: I played a little with CMake imports and those files also have been added.

Xavrax:
> Would you consider incorporating this into espressif’s [Component Registry](https://components.espressif.com/)? Their guide is provided [here](https://docs.espressif.com/projects/idf-component-manager/en/latest/guides/packaging_components.html).
That's a great idea. I will take a look into that and I will try to satisfy all the needs required by the registry.

Xavrax:
@blake-spangenberg 
We've asked our legals if we're okay with this 3rd party service. 
I will let you know about the decision as soon as it will be made. 

Until then you will need to rawly `git clone` the repository. 
(hopefully they will allow us to publish our repo there and the SDK will be available at the registry to make the developer experience at the highest level - we've made same thing for rust's SDK so it shouldn't be a problem).

CMakeLists.txt
blake-spangenberg: I'm still getting some compilation failures due to compiler warnings:

```
/home/ubuntu/git/hub-3/c-core/core/pbcc_set_state.c: In function 'pbcc_adjust_state':
/home/ubuntu/git/hub-3/c-core/core/pbcc_set_state.c:101:17: error: 'strncpy' output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
  101 |                 strncpy(curr_cg, str_cg, cg_len);
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/ubuntu/git/hub-3/c-core/core/pbcc_set_state.c:95:61: note: length computed here
   95 |                 if (cg_temp == NULL) { end = true; cg_len = strlen(str_cg); }
      |                                                             ^~~~~~~~~~~~~~
/home/ubuntu/git/hub-3/c-core/core/pubnub_ccore_pubsub.c: In function 'pbcc_set_user_id':
/home/ubuntu/git/hub-3/c-core/core/pubnub_ccore_pubsub.c:202:9: error: 'strncpy' output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
  202 |         strncpy(pb->user_id, user_id, pb->user_id_len);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/ubuntu/git/hub-3/c-core/core/pubnub_ccore_pubsub.c:187:47: note: length computed here
  187 |     pb->user_id_len = (NULL == user_id) ? 0 : strlen(user_id);
      |                                               ^~~~~~~~~~~~~~~
```

Consider using `strlcpy` instead of `strncpy` in these use cases. [Here's a nice summary of why you would want to do this](https://www.geeksforgeeks.org/why-strcpy-and-strncpy-are-not-safe-to-use/)

Xavrax: One more thing. 
Is it the last error message that you've received? 

I'm asking because I see that we got totally different environments and my `-Wall` doesn't see the issues that yours points out. 

I don't want to ask for each error that you get because I don't want to overwhelm you... :/
If there are any other errors - can you describe me your environment? OS, compiler, linter (if any) and flags used?
It will help me to find all the issues that you're struggling with during the compilation process and let me fix all of them before the release. ;d

blake-spangenberg: @Xavrax 

## Regarding `strncmp` warning

While it's ultimately your discretion as to how to fix the warning, I would suggest simply downgrading to `strcpy` from `strncpy` in the two problem scenarios. For example:

```c
char* curr_cg = (char*)malloc(cg_len + 1);
strncpy(curr_cg, str_cg, cg_len);
curr_cg[cg_len] = '\0';
```

becomes...

```c
char* curr_cg = malloc(cg_len + 1);
strcpy(curr_cg, str_cg);
```

This is safe because:

1. You're allocated enough room for the original string and it's null-terminator
2. You already assume `str_cg` is null-terminated (via your earlier `strlen` call)
3. `strcpy` will copy the null-terminator from `str_cg` to `curr_cg`

If done in both places listed above, it would resolve our compilation errors and remain c89 compatible without sacrificing system stability.

*These two are the only two blocking compilation errors I'm encountering when attempting to build. It throws some 5 other warnings, but those do not stop us from compiling.*

## Regarding Build Environment

Here are some details of how we build our project:

```shell
$ ~/.espressif/tools/xtensa-esp32-elf/esp-2021r1-8.4.0/xtensa-esp32-elf/bin/xtensa-esp32-elf-gcc --version
xtensa-esp32-elf-gcc (crosstool-NG esp-2021r1) 8.4.0

$ cmake --version
cmake version 3.22.1

$ cd esp-idf && git describe --tags
v5.2.2
```

```shell
$ cat CMakeLists.txt
```

```cmake
cmake_minimum_required(VERSION 3.16)
set(EXTRA_COMPONENT_DIRS "main" "${CMAKE_CURRENT_SOURCE_DIR}/c-core")
set(PROJ_NAME hub3)
set(ENV{IDF_TARGET} "esp32")
add_definitions(-DMBEDTLS="ON")
set(ENV{MBEDTLS} "ON")
include($ENV{IDF_PATH}/tools/cmake/project.cmake)
project(${PROJ_NAME})
```

```shell
$ cat sdkconfig | grep COMPILER
CONFIG_BOOTLOADER_COMPILER_OPTIMIZATION_SIZE=y
# CONFIG_BOOTLOADER_COMPILER_OPTIMIZATION_DEBUG is not set
# CONFIG_BOOTLOADER_COMPILER_OPTIMIZATION_PERF is not set
# CONFIG_BOOTLOADER_COMPILER_OPTIMIZATION_NONE is not set
# CONFIG_COMPILER_OPTIMIZATION_DEBUG is not set
# CONFIG_COMPILER_OPTIMIZATION_SIZE is not set
CONFIG_COMPILER_OPTIMIZATION_PERF=y
# CONFIG_COMPILER_OPTIMIZATION_NONE is not set
CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_ENABLE=y
# CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_SILENT is not set
# CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE is not set
CONFIG_COMPILER_FLOAT_LIB_FROM_GCCLIB=y
CONFIG_COMPILER_OPTIMIZATION_ASSERTION_LEVEL=2
# CONFIG_COMPILER_OPTIMIZATION_CHECKS_SILENT is not set
CONFIG_COMPILER_HIDE_PATHS_MACROS=y
# CONFIG_COMPILER_CXX_EXCEPTIONS is not set
# CONFIG_COMPILER_CXX_RTTI is not set
CONFIG_COMPILER_STACK_CHECK_MODE_NONE=y
# CONFIG_COMPILER_STACK_CHECK_MODE_NORM is not set
# CONFIG_COMPILER_STACK_CHECK_MODE_STRONG is not set
# CONFIG_COMPILER_STACK_CHECK_MODE_ALL is not set
# CONFIG_COMPILER_WARN_WRITE_STRINGS is not set
# CONFIG_COMPILER_DISABLE_GCC12_WARNINGS is not set
# CONFIG_COMPILER_DISABLE_GCC13_WARNINGS is not set
# CONFIG_COMPILER_DUMP_RTL_FILES is not set
CONFIG_COMPILER_RT_LIB_GCCLIB=y
CONFIG_COMPILER_RT_LIB_NAME="gcc"
# CONFIG_COMPILER_OPTIMIZATION_LEVEL_DEBUG is not set
# CONFIG_COMPILER_OPTIMIZATION_DEFAULT is not set
# CONFIG_COMPILER_OPTIMIZATION_LEVEL_RELEASE is not set
```

Xavrax: @blake-spangenberg 
> While it's ultimately your discretion as to how to fix the warning,
Yeah, but i wanted to fit best into your needs. 

> , I would suggest simply downgrading to strcpy from strncpy in the two problem scenarios.
If it will suspend your errors - i will follow this way. 

> Regarding Build Environment
Thank you, I will try to configure my testing env to the same values and check if it will compile before pushing.

Xavrax: @blake-spangenberg 
I've replaced all occurrences of `strncpy` with the `strcpy` function as you suggested. I've read the articles about the differences between `strn`, `strl`, and `str` functions, and to be honest, `strncpy` can be useful in some situations, but it also introduces potential risks and people tend to avoid it. 

Btw. our settings seem very similar, so I'm not sure why I haven't encountered any compilation issues. (I'm using a slightly newer version of GCC and CMake, but I suspect that this kind of issue could be related to the older compiler, not the never one).

Please let me know if this change fixes all of your compilation problems.
I've noticed that you work at 8 pm my time, so I'll try to be online to resolve any other issues you find as quickly as possible. I'd like to release it today, but only if it works on your end.

Thanks for your patience and willingness to help!

core/pubnub_generate_uuid.c
```diff
@@ -46,7 +46,11 @@ int pubnub_generate_uuid_v1_time(
     ud->clock_seq_low = *io_clock_seq & 0xFF;
     ud->clock_seq_hi_and_reserved = (*io_clock_seq & 0x3F00) >> 8;
     ud->clock_seq_hi_and_reserved |= 0x80;
+#ifdef ESP_PLATFORM
+    memcpy(ud->node, i_node, sizeof ud->node);
```
parfeon, line 50: Since there is a conditional macro, it looks like there is some difference in `memcpy` interface for `ESP` and other platforms. Do we pass values, not pointers?

freertos/pubnub_config.h
```diff
@@ -70,7 +70,7 @@
     in milliseconds. Timeout duration in the context can be changed by the 
     user after initialization.
     */
-#define PUBNUB_DEFAULT_TRANSACTION_TIMER    310000
+#define PUBNUB_DEFAULT_TRANSACTION_TIMER    1024
```
parfeon, line 73: Why did we reduce transactional request timeout (subscribe)?

lib/sockets/pbpal_handle_socket_error.c
```diff
@@ -31,7 +31,9 @@ void pbpal_report_error_from_environment(pubnub_t* pb, char const* file, int lin
         errno,
         err_str);
     if (pb != NULL) {
+#if PUBNUB_BLOCKING_IO_SETTABLE
         PUBNUB_LOG_DEBUG(" use_blocking_io=%d\n", (int)pb->options.use_blocking_io);
+#endif
```
parfeon, line 36:
```suggestion
#else
        PUBNUB_LOG_DEBUG("\n");
#endif // PUBNUB_BLOCKING_IO_SETTABLE
```

mbedtls/pbpal_connect_mbedtls.c
```diff
@@ -0,0 +1,256 @@
...
+    "yj1hTdNGCbM+w6DjY1Ub8rrvrTnhQ7k4o+YviiY776BQVvnGCv04zcQLcFGUl5gE\n"
+    "38NflNUVyRRBnMRddWQVDf9VMOyGj/8N7yy5Y0b2qvzfvGn9LhJIZJrglfCm7ymP\n"
+    "AbEVtQwdpf5pLGkkeB6zpxxxYu7KyJesF12KwvhHhm4qxFYxldBniYUr+WymXUad\n"
+    "DKqC5JlR3XC321Y9YeRq4VzW9v493kHMB65jUr9TU/Qr6cf9tveCX4XSQRjbgbME\n"
+    "HMUfpIBvFSDJ3gyICh3WZlXi/EjJKSZp4A==\n"
+    "-----END CERTIFICATE-----\n";
+
+static void alloc_setup(pubnub_t* pb);
+
+static const char* get_origin(pubnub_t* pb)
```
parfeon, line 86: In DNS resolution implementation we have similar condition to pick `host` which should be resolved. Can we just use `get_origin(pb)` function there instead?

parfeon (approved):
LGTM!
---
_There are few small questions._

mbedtls/pbpal_connect_mbedtls.c
Xavrax: Sure, probably it's left over of the initial concept of the mbedtls module. I will fix that.

freertos/pubnub_config.h
Xavrax: Leftover after testing. Gonna return to the previous value

core/pubnub_generate_uuid.c
Xavrax: Good catch. I don't know why I've made that change. 
I will change back that into previous value.

CMakeLists.txt
Xavrax: I think that I resolved the issues. If there are any more - let us know, we will fix them as soon as possible in the next PR.

Xavrax:
@pubnub-release-bot release

blake-spangenberg:
Can confirm publish/subscribe is working on our esp32-based hardware platform. Thanks for the hard work everyone!

We'll be in touch if there's any problems or follow-up questions.

Xavrax:
Thank you for the confirmation and have fun with our SDK <3

PR #185, opened 2024-06-07
strings in objects API

Xavrax:
feat: possibility to use string in actions API

Added possibility to use strings in actions API

refactor: `pubnub_action_type` deprecated

`pubnub_action_type` enum has been deprecated

### DIFF
--- a/core/pbcc_actions_api.c
+++ b/core/pbcc_actions_api.c
@@ -23,10 +23,37 @@ enum pubnub_res pbcc_form_the_action_object(struct pbcc_context* pb,
                                             char* obj_buffer,
                                             size_t buffer_size,
                                             enum pubnub_action_type actype,
-                                            char const** val)
-{
-    char const* user_id = pbcc_user_id_get(pb);
+                                            char const** val) {
     char const* type_literal;
+
+    switch(actype) {
+    case pbactypReaction:
+        type_literal = "reaction";
+        break;
+    case pbactypReceipt:
+        type_literal = "receipt";
+        break;
+    case pbactypCustom:
+        type_literal = "custom";
+        break;
+    default:
+        PUBNUB_LOG_ERROR("pbcc_form_the_action_object(pbcc=%p) - "
+                         "unknown action type = %d\n",
+                         pb,
+                         actype);
+        return PNR_INVALID_PARAMETERS;
+    }
+
+    return pbcc_form_the_action_object_str(pb, obj_buffer, buffer_size, type_literal, val);
+}
+
+
+enum pubnub_res pbcc_form_the_action_object_str(struct pbcc_context* pb,
+                                            char* obj_buffer,
+                                            size_t buffer_size,
+                                            char const* action_type,
+                                            char const** val) {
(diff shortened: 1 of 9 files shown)


core/pbcc_actions_api.c
```diff
@@ -17,31 +17,16 @@
 
 /* Maximum number of actions to return in response */
 #define MAX_ACTIONS_LIMIT 100
+#define MAX_ACTION_VALUE_LENGTH 15
```
parfeon, line 20:
```suggestion
#define MAX_ACTION_TYPE_LENGTH 15
```
Should it be like this?

core/pbcc_actions_api.c
```diff
@@ -59,8 +44,35 @@ enum pubnub_res pbcc_form_the_action_object(struct pbcc_context* pb,
...
+        return PNR_INVALID_PARAMETERS;
+    }
+    if (('\"' != **val) || ('\"' != *(*val + pb_strnlen_s(*val, PUBNUB_MAX_OBJECT_LENGTH) - 1))) {
+        PUBNUB_LOG_ERROR("pbcc_form_the_action_object(pbcc=%p) - "
+                         "quotation marks on value ends are missing: "
+                         "value = %s\n",
+                         pb,
+                         *val);
+        return PNR_INVALID_PARAMETERS;
+    }
```
parfeon, line 72: If users now able to pass `string` as `type`, should we have the same check for it and ask users to enclose `type` into double quotes?

parfeon (approved):
Looks good but still some questions :)

core/pbcc_actions_api.c
Xavrax: tbh, we should - commiting

core/pbcc_actions_api.c
Xavrax: But wait. 
previous implementation haven't include the quotes. So in this case they shouldn't be forced to do such a thing ;o

core/pbcc_actions_api.c
```diff
@@ -70,6 +70,14 @@ enum pubnub_res pbcc_form_the_action_object_str(struct pbcc_context* pb,
                          *val);
         return PNR_INVALID_PARAMETERS;
     }
+    if (('\"' != *action_type) || ('\"' != *(action_type + pb_strnlen_s(action_type, PUBNUB_MAX_OBJECT_LENGTH) - 1))) {
+        PUBNUB_LOG_ERROR("pbcc_form_the_action_object(pbcc=%p) - "
+                         "quotation marks on value ends are missing: "
```
parfeon, line 75: `"quotation marks on value ends are missing: "` -> `"quotation marks on action type ends are missing: "`

parfeon (approved):
LGTM!

Xavrax:
@pubnub-release-bot release

Xavrax:
@pubnub-release-bot release

parfeon:
@pubnub-release-bot release

parfeon:
@pubnub-release-bot release as v4.10.0

PR #188, opened 2024-06-28
fix cmake file for not ESP32 builds

Xavrax:
fix: fix `cmake` build for not ESP32 builds

Fixed `cmake` build for not ESP32 builds


seba-aln (approved):
Extremely well done work! You can see the years at university finally paying off!
Love it!

Xavrax:
@pubnub-release-bot release

PR #191, opened 2024-08-02
Fix history query parameter values

parfeon:
fix(history): fix query parameter values

Fix query values for boolean flags for history endpoint (`include meta`, `include uuid`, `include message type` and `reverse`).


Xavrax (approved):

parfeon:
@pubnub-release-bot release

PR #193, opened 2024-08-05
fix custom `bool` type for CMake builds

Xavrax:
fix: fix custom `bool` type for CMake builds

fixed custom `bool` type for CMake builds


seba-aln (commented):
ok

PR #195, opened 2024-09-04
fix not needed null character

Xavrax:
fix: removed additional null byte character

removed additional null byte character


parfeon (approved):
LGTM

Xavrax:
@pubnub-release-bot release

PR #196, opened 2024-09-11
Subscription Event Engine and Events Listener implementation

parfeon:
feat(event-engine): add general event engine

Add core Event Engine implementation with the required set of types and methods.

feat(subscribe-event-engine): add subscribe event engine

Add Subscribe Event Engine built atop of the core Event Engine implementation.

feat(entities): add base PubNub entities

Add the following entities: channel, channel group, uuid and channel metadata objects.

feat(subscription): add subscription and subscription sets

Add objects to manage subscriptions and provides interface for update listeners.

feat(listeners): add event listener

Add new event listeners, which make it possible to add listeners to a specific entity or group of entities (though subscription and subscription set).

feat(retry): add request retry configuration

Added ability to configure automated retry policies for failed requests.


Xavrax:
[image
](https://tenor.com/bLuSE.gif)
10k changes

Xavrax (commented):
18/40 files reviewed. 

Overall I love your design. 
Some questions. 

Also it would be nice to have example that shows users how to use the new API.

core/pubnub_internal_common.h
```diff
@@ -3,6 +3,7 @@
 #if !defined INC_PUBNUB_INTERNAL_COMMON
 #define INC_PUBNUB_INTERNAL_COMMON
 
+#include <stdbool.h>
```
Xavrax, line 6: I see many random import changes. 
Probably they're related to the architecture and feature sets. Am I right?

Xavrax (commented):
Seems good (30/40 files review)

.github/workflows/run-tests.yml
```diff
@@ -21,8 +21,8 @@ jobs:
         runner:
           - os: ubuntu-latest
             group: Default
-          - os: macos-13
-            group: macos-gh
+#          - os: macos-13
+#            group: macos-gh
```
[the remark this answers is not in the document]
parfeon: I will check it with master before release :)

core/pbcc_event_engine.c
```diff
@@ -0,0 +1,468 @@
...
+#include "core/pubnub_assert.h"
+#include "lib/pbref_counter.h"
+#include "core/pubnub_log.h"
+
+
+// ----------------------------------------------
+//                     Types
+// ----------------------------------------------
+
+// Sharable data type definition.
```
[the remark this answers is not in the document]
parfeon: I tried to find a reason to use this form, but only multiline comments (which is not the case for one-liner) came to my mind. I don't mind. Will change.

core/pbcc_event_engine.c
```diff
@@ -0,0 +1,468 @@
...
+    enum pubnub_res  rslt        = PNR_OK;
+    const pbarray_t* invocations = transition->invocations;
+    const size_t     count       = pbarray_count(invocations);
+    for (size_t i = 0; i < count; ++i) {
+        pbcc_ee_invocation_t* inv = _pbcc_ee_invocation_copy(
+            (pbcc_ee_invocation_t*)pbarray_element_at(invocations, i));
+        if (PBAR_OK != pbarray_add(ee->invocations,inv)) {
+            pbcc_ee_invocation_free(inv);
+            rslt = PNR_OUT_OF_MEMORY;
+        }
```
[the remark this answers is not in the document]
parfeon: Good catch!
`break` is missing and failed transition invocation handled few lines below by removing invocations which have been added from this new transition into EE invocations queue.

core/pubnub_internal_common.h
parfeon: Some imports have been removed because IDE shown that they are unused, some added in other place.

To be honest, `<stdbool.h>` and `core/c99/stdbool.h`: not sure about intention. Do we still need the latest c89 support (as I understand this is the only reason why `core/c99/stdbool.h` exists because `bool` is missing from C89 std)?

Moreover, additionally I was confused by feature flags which let specify custom booleans at a compile-time and wasn't sure whether it needed to be utilized in any way.

Xavrax: >  Do we still need the latest c89 support 
This topic is still under the discussion. :/

lib/pbarray.c
```diff
@@ -0,0 +1,643 @@
+/* -*- c-file-style:"stroustrup"; indent-tabs-mode: nil -*- */
```
Xavrax, line 1: Tbh - I would love to have such utilities all across the SDK ;d

core/pubnub_subscribe_event_engine.c
```diff
@@ -0,0 +1,802 @@
...
+            sub,
+            sub->entity->id.ptr);
+        pubnub_mutex_unlock(set->mutw);
+        return PNR_SUB_NOT_FOUND;
+    }
+
+    if (subscribed) {
+        // Temporarily release lock to avoid deadlocks from subscription ee core
+        // code access the list of subscribables.
+        pubnub_mutex_unlock(set->mutw);
```
Xavrax, line 304: I'm a little bit worried about this statement. 
Is it possible that this quite short amount of time unguarded mutex might mess anything in execution?

core/pbcc_subscribe_event_listener.c
```diff
@@ -0,0 +1,660 @@
...
+    }
+
+    return PNR_OK;
+}
+
+void _pbcc_listener_free(pbcc_listener_t* listener)
+{
+    if (NULL == listener) { return; }
+    if (0 == pbref_counter_free(listener->counter)) { free(listener); }
+}
```
Xavrax, line 612: Shouldn't we set `listener` to `NULL`?

core/pbcc_subscribe_event_engine.c
```diff
@@ -0,0 +1,2485 @@
+#include "pbcc_subscribe_event_engine.h"
```
Xavrax, line 1: Nitpicking: If I were you, I would split this file into smaller ones. ;d

Xavrax (commented):
Overall LGTM

Some new comments.

+ I'm asking for sub EE sample to progress with the PR ;d.

.clang-format
```diff
@@ -88,7 +88,7 @@ PenaltyExcessCharacter: 4
 PenaltyReturnTypeOnItsOwnLine: 100
```
Xavrax, line 88: Do you use format for single files? 
I want to reformat the whole repository but there are some issues with compilation so I'm confused how does it work on your side.

CMakeLists.txt
```diff
@@ -60,7 +60,9 @@ num_option(ONLY_PUBSUB_API "Only pubsub API" OFF)
 num_option(USE_PROXY "Use proxy" ON)
 num_option(USE_GZIP_COMPRESSION "Use gzip compression" ON)
 num_option(RECEIVE_GZIP_RESPONSE "Use gzip decompression" ON)
+num_option(USE_RETRY_CONFIGURATION "Use requests retry" ON)
 num_option(USE_SUBSCRIBE_V2 "Use subscribe v2" ON)
+num_option(USE_SUBSCRIBE_EVENT_ENGINE "Use Subscribe Event Engine" ON)
```
Xavrax, line 65: Does this flag influent current users of the SDK?

core/pbauto_heartbeat.c
```diff
@@ -59,6 +59,7 @@ static int copy_context_settings(pubnub_t* pb_clone, pubnub_t const* pb)
     PUBNUB_ASSERT_OPT(pb_valid_ctx_ptr(pb));
 
     pubnub_mutex_lock(pb_clone->monitor);
+    pb_clone->core.auth_token = pb->core.auth_token;
```
Xavrax, line 62: Why is this line needed?

core/pbcc_memory_utils.c
```diff
@@ -0,0 +1,17 @@
...
+//                   Functions
+// ----------------------------------------------
+
+void pbcc_free_ptr(void** ptr)
+{
+    if (NULL == ptr || NULL == *ptr) return;
+
+    free(*ptr);
+    *ptr = NULL;
+}
```
Xavrax, line 17: This is quite risky approach. 

What in case whereas `ptr` points to the allocated memory. 
I know that propably not in your use-case but I'm aware of the future development and using this function on allocated pointer to allocated pointer (which is ofc very tricky and nobody should do this ofc) might fail. 

I can assume some array of pointers that might be failing here.

core/pbcc_memory_utils.h
```diff
@@ -0,0 +1,61 @@
...
+ */
+#define PBCC_ALLOCATE_TYPE(var, type, print_error_message, return_value) \
+    type *var = (type *)malloc(sizeof(type));                            \
+    if (NULL == var) {                                                   \
+        if (true == print_error_message) {                               \
+            PUBNUB_LOG_ERROR("[%s] Failed to allocate memory for '%s'\n",\
+                             __func__, #type);                           \
+        }                                                                \
+        return return_value;                                             \
+    }
```
Xavrax, line 48: I'm aware of macros. 
This one seems to be valid but I'm affraid that it might generate very strange compilation errors.

core/pbcc_memory_utils.h
```diff
@@ -0,0 +1,61 @@
...
+ * @endcode
+ *
+ * @param var                 The name of the variable which will be assigned to
+ *                            the allocated memory pointer.
+ * @param type                Name of type for which memory will be allocated.
+ * @param print_error_message Whether a memory allocation error should be
+ *                            printed to the console and log file or not.
+ * @param return_value        What should be returned in case of a memory
+ *                            allocation error. The value can be 'NULL' or
+ *                            'enum' field.
```
Xavrax, line 38: There should be information that it will return the value from the current function because it might leads to misunderstanding how this macro works.

core/pbcc_request_retry_timer.c
```diff
@@ -0,0 +1,198 @@
...
+//                   Constants
+// ----------------------------------------------
+
+/**
+ * @brief Polling timeout.
+ *
+ * Timeout used with sleep function inside of `while` loop to break constant
+ * polling and give other processes some CPU time.
+ */
+#define POLLING_TIMEOUT 10
```
Xavrax, line 26: Risky define. 
If anyone would import that we can redefine their definitions. 
```suggestion
#define PBCC_POLLING_TIMEOUT 10
```

openssl/posix.mk
```diff
@@ -109,7 +109,12 @@ REVOKE_TOKEN_SOURCEFILES = ../core/pubnub_revoke_token_api.c ../core/pbcc_revoke
 REVOKE_TOKEN_OBJFILES = pubnub_revoke_token_api.o pbcc_revoke_token_api.o
 endif
 
-CFLAGS = -g -D PUBNUB_THREADSAFE -D PUBNUB_LOG_LEVEL=PUBNUB_LOG_LEVEL_WARNING -Wall -D PUBNUB_ONLY_PUBSUB_API=$(ONLY_PUBSUB_API) -D PUBNUB_PROXY_API=$(USE_PROXY) -D PUBNUB_USE_GZIP_COMPRESSION=$(USE_GZIP_COMPRESSION) -D PUBNUB_RECEIVE_GZIP_RESPONSE=$(RECEIVE_GZIP_RESPONSE) -D PUBNUB_USE_SUBSCRIBE_V2=$(USE_SUBSCRIBE_V2) -D PUBNUB_USE_OBJECTS_API=$(USE_OBJECTS_API) -D PUBNUB_USE_ACTIONS_API=$(USE_ACTIONS_API) -D PUBNUB_USE_AUTO_HEARTBEAT=$(USE_AUTO_HEARTBEAT) -D PUBNUB_USE_GRANT_TOKEN_API=$(USE_GRANT_TOKEN) -D PUBNUB_USE_REVOKE_TOKEN_API=$(USE_REVOKE_TOKEN) -D PUBNUB_USE_FETCH_HISTORY=$(USE_FETCH_HISTORY) 
+ifeq ($(USE_CRYPTO_API), 1)
```
Xavrax, line 116: This line renders on gh with a very strange way. 
Probably there are a lot of whitespaces at the end of the line.

lib/pbstrdup.c
```diff
@@ -0,0 +1,55 @@
...
+ * @param src       Pointer to the null-terminated byte string to copy.
+ * @param len       How many (at most) bytes from `src` should be written into
+ *                  copy.
+ * @param check_len Whether `len` should be checked to not be larger that `src`
+ *                  byte string length.
+ * @return Pointer to the null-terminated byte string which is copied from
+ *         `src`. `NULL` value will be returned in case of insufficient memory
+ *         error.
+ */
+static char* _pbstrndup(const char* src, size_t len, bool check_len);
```
Xavrax, line 24: Question:
Do we want to document private stuff? 

I'm not against but it would be nice to have a common way how do we treat them.

core/pubnub_netcore.c
```diff
@@ -1090,6 +1098,15 @@ int pbnc_fsm(struct pubnub_* pb)
...
             break;
+        case PNR_TX_BUFF_TOO_SMALL:
+            /** We could copy the "line so far" to the reply buffer
+                and then process the line from the reply buffer when
+                it's finished. This would be much more complex, yet
+                most lines will not be that long and, if the reply
+                buffer is static, it, too, might not be large enough.
+            */
+            pb->state = PBS_RX_HEADERS;
+            goto next_state;
```
Xavrax, line 1109: Will this affect current users? 
As I understood we're changing the behaviour of the whole SDK in case of the buffer overflow.

core/pubnub_entities.c
```diff
@@ -0,0 +1,154 @@
...
+        pb,
+        id,
+        PUBNUB_ENTITY_USER_METADATA);
+}
+
+bool pubnub_entity_free(void** entity)
+{
+    if (NULL == entity || NULL == *entity) { return false; }
+
+    printf("~~~~~~> RESULT: %d\n", is_pubnub_entity_(*entity));
```
Xavrax, line 89: debug print

core/samples/subscribe_event_engine_sample.c
```diff
@@ -0,0 +1,28 @@
+/* -*- c-file-style:"stroustrup"; indent-tabs-mode: nil -*- */
```
Xavrax, line 1: Have you used sanitizers to test it? 

If not I can do it for you tomorrow (or any other day).

core/pbcc_subscribe_event_engine_events.c
```diff
@@ -0,0 +1,333 @@
...
+    const pbcc_subscribe_ee_event type = SUBSCRIBE_EE_EVENT_DISCONNECT;
+    return pbcc_subscribe_ee_event_alloc_(ee, type, NULL, NULL, NULL, NULL);
+}
+
+pbcc_ee_event_t* pbcc_reconnect_event_alloc(
+    pbcc_subscribe_ee_t*            ee,
+    const pubnub_subscribe_cursor_t cursor)
+{
+    const pbcc_subscribe_ee_event type = SUBSCRIBE_EE_EVENT_RECONNECT;
+    return pbcc_subscribe_ee_event_alloc_(ee, type, NULL, NULL, &cursor, NULL);
```
Xavrax, line 195: Conclusion, not a review note:
When I see the C code I would really love to see builders... 
There is a lot of `NULL` statements that means nothing during the code review. 

image

core/pubnub_retry_configuration.c
```diff
@@ -0,0 +1,470 @@
...
+
+    return configuration;
+}
+
+pubnub_retry_configuration_t*
+pubnub_retry_configuration_linear_alloc_with_options(
+    const int delay,
+    const int maximum_retry,
+    const int excluded,
+    ...)
```
Xavrax, line 207: Do you know what compilers are supporting this syntax? 

I don't know and I'm thinking if any of our current customers might be influenced by this change and make their build fails.

Xavrax (commented):
There was a lot of files and lines. 
I hope that I didn't miss anything. 

Overall LGTM.
Some questions and things to do before the merge.

core/pbcc_subscribe_event_engine_events.c
parfeon: IDE will show docs for those arguments, so it will have a bit more meaning :)

.clang-format
parfeon: Hm, I didn't receive a penalty there.
Also, it already was there and the only thing which I changed was the `Standard:` `value: `Cpp11` → `c++11` (previous value wasn't recognized)

CMakeLists.txt
parfeon: This is temporary while was developing or IDE won't highlight the code properly :/ 
By default, Subscribe EE and retry will be off.

core/pbcc_memory_utils.c
parfeon: I will remove this function because the task for which it has been done has been solved in another way :)

core/pbcc_memory_utils.h
parfeon: What do you mean by “strange”?

lib/pbstrdup.c
parfeon: Usually, we work with private functions, and occasionally, it is not obvious what those arguments are expected to be and what the function does.

“Common” way - do you mean document all functions? I'm trying to keep to the approach where I document new code or code which I rewrite.

core/pubnub_netcore.c
parfeon: Where it came from? oO

core/pubnub_retry_configuration.c
parfeon: This is the same notation as `printf` has, which I believe supported in most compilers which support the standard C library.

core/pbcc_memory_utils.h
Xavrax: I'm back in the days when compilers couldn't distinguish the generated from macro code from the hand-written ones. 
I remember that I had to count lines of macro to find where exactly the problem occurred ;D

But I'm not against macros - just wanted to add a note about it. ;d

.clang-format
Xavrax: mhm. 
Please do not format whole codebase within this PR. 

I want to do it at some point, but it makes ~80k lines change :D

core/pbcc_request_retry_timer.c
```diff
@@ -0,0 +1,198 @@
...
+    pbcc_request_retry_timer_stop(timer);
+
+    pubnub_mutex_lock(timer->mutw);
+    timer->running = true;
+    timer->delay   = delay;
+    pubnub_mutex_unlock(timer->mutw);
+    pubnub_mutex_lock(timer->pb->monitor);
+    timer->pb->state = PBS_WAIT_RETRY;
+    timer->pb->core.http_retry_count++;
+    pubnub_mutex_unlock(timer->pb->monitor);
```
Xavrax, line 123: Q: is it okay to lock those locks separetely?
Won't it cause any UB?

Xavrax (commented):
Great job! 
LGTM

Few questions. 
Gonna approve right after sanitizers' issues are resolved.

microchip_harmony/pubnub_ntf_harmony.c
```diff
@@ -148,6 +153,27 @@ void pbntf_lost_socket(pubnub_t *pb, pb_socket_t socket)
...
+            }
+        }
+    }
+
+    /** There were no need to start retry timer, we can free it if exists. */
+    if (NULL != pb->core.retry_timer) {
+        pb->core.http_retry_count = 0;
+        pbcc_request_retry_timer_free(&pb->core.retry_timer);
+    }
+#endif // #if PUBNUB_USE_RETRY_CONFIGURATION
```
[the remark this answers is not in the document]
parfeon: @Xavrax can you elaborate more on this one? This code is identical in three files - do you want to add new function which have this code inside and use it in those files (function will return boolean to know whether flow should continue or not)?

openssl/posix.mk
```diff
@@ -51,7 +55,7 @@ USE_FETCH_HISTORY = 1
 endif
 
 ifndef USE_CRYPTO_API
-USE_CRYPTO_API = 0
+USE_CRYPTO_API = 1
 endif
```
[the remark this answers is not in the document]
parfeon: @Xavrax in [current file](https://github.com/pubnub/c-core/blob/master/openssl/posix.mk#L1) (at the end of the line) you can see that crypto files were there unconditionally and when gurads has been fixed it started to cause problems. 
Here it has been enabled by default because CI runs `all` task which require crypto module to be used - this is the reason why I enabled it in those files.

core/pbcc_subscribe_v2.c
```diff
@@ -328,8 +328,8 @@ struct pubnub_v2_message pbcc_get_msg_v2(struct pbcc_context* p)
     }
 
     if (jonmpOK == pbjson_get_object_value(&el, "b", &found)) {
-        rslt.match_or_group.ptr  = (char*)found.start;
-        rslt.match_or_group.size = found.end - found.start;
+        rslt.match_or_group.ptr  = (char*)found.start + 1;
+        rslt.match_or_group.size = found.end - found.start - 2;
```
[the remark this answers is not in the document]
parfeon: @Xavrax When I tried to find listener for group, I've got something like this: `"channel_group\"`. Then I checked how regular channel parsed, and it looks like there were these fixed added before.

core/pbcc_request_retry_timer.c
parfeon: @Xavrax I don't think it is critical, but I can put one of them inside another after sanitizer check (I hope it is not critical).

parfeon:
@pubnub-release-bot release

PR #199, opened 2024-11-14
Add custom message type support

parfeon:
feat(message-type): add custom message type support

Add custom message type support for the following APIs: publish, signal, share file, subscribe and history.

feat(ipv6): add connectivity setter
    
Add `pubnub_set_ipv4_connectivity` and `pubnub_set_ipv6_connectivity` to `pubnub_coreapi` to switch preferred connectivity protocol.
    
fix(proxy): add proxy context reset on connection close
    
Make sure that in case of connection close (including because of error) proxy context object will be reset.


core/pubnub_coreapi_ex.c
```diff
@@ -27,12 +27,13 @@
...
-    result.replicate  = true;
-    result.meta       = NULL;
-    result.method     = pubnubSendViaGET;
-    result.ttl        = SIZE_MAX;
+    result.store               = true;
+    result.cipher_key          = NULL;
+    result.replicate           = true;
+    result.meta                = NULL;
+    result.method              = pubnubSendViaGET;
+    result.ttl                 = SIZE_MAX;
```
KGronek-Pubnub, line 35: In pubnub_publish_defopts method description there is mention of all default values, except tll and custom_message_type. I think it could be worth to add those as well.

parfeon: Sure, we can do it :)

KGronek-Pubnub (approved):
Everything looks good to me, compiles well on Windows :)

parfeon:
@pubnub-release-bot release

PR #201, opened 2025-01-03
Add logging callback

KGronek-Pubnub:
feat(log): add possibility to replace default log function with a callback provided by the user


core/pubnub_log.h
```diff
@@ -35,18 +36,36 @@ enum pubnub_log_level {
...
 #if !defined PUBNUB_LOG_PRINTF
 #include <stdio.h>
 /** User should define a printf-like function that will do the actual
     logging. If it is not defined, we'll use printf().
 */
-#define PUBNUB_LOG_PRINTF(...) printf(__VA_ARGS__)
+#define PUBNUB_LOG_PRINTF(LVL, ...) \
+    do { \
+        if (pubnub_log_callback) { \
+            char logMessage[1024]; \
```
parfeon, line 56: @KGronek-Pubnub Buffer size won't be an issue here?
Also, I'm concerned about change in signature because of [this](https://github.com/pubnub/c-core/blob/master/microchip_harmony/pubnub_callback.h#L52).

KGronek-Pubnub: @parfeon yeah I was afraid that this change in signature could cause some potential issues :/ thanks
I will change this logic to happen in PUBNUB_LOG macro instead.
As for the buffer size, I can change to calculate it dynamically.

core/pubnub_log.h
```diff
@@ -44,9 +58,21 @@ enum pubnub_log_level {
...
-    interface
+    interface. Uses callback instead of printf if pubnub_log_callback
+    is set.
 */
-#define PUBNUB_LOG(LVL, ...) do { if (LVL <= PUBNUB_LOG_LEVEL) PUBNUB_LOG_PRINTF(__VA_ARGS__); } while(0)
+#if !defined PUBNUB_LOG
+#define PUBNUB_LOG(LVL, ...) \
+    do { \
+        if (LVL <= PUBNUB_LOG_LEVEL) { \
+            if (pubnub_log_callback) { \
```
Xavrax, line 68: Are we sure that we want to have this dynamic decision making? 
Shouldn't we make it a compilation flag? 

I mean it makes sense when a flag would be set to TRUE (to check if somebody added the callback) but it makes a difference as it is called very frequently for our customers that doesn't need this feature. 
Probably the performance loss won't be even noticed but I want to start the discussion about it.

Xavrax: @KGronek-Pubnub @parfeon 
What are your opinions?

Xavrax:
Anyway overall LGTM - just discuss and resolve the conversation.

core/pubnub_log.h
KGronek-Pubnub: Sure I could rework it to be a flag instead, just I think there are already so many compilation flags, so for end users it might be very difficult to understand and compile the version they need.
I don't think validation check can make any performance loss, so it's more a design decision here.

Xavrax (approved):
LGTM

KGronek-Pubnub:
@pubnub-release-bot release

PR #206, opened 2025-02-11
Fix truncated `signature` query value

parfeon:
fix(crypto): fix truncated `signature` query value

Fix issue because of which `signature` value in query has been truncated.


parfeon:
@pubnub-release-bot release

PR #207, opened 2025-02-19
introduce `void*` parameter for user data in listeners

Xavrax:
feat: introduce `void*` parameter for user data in listeners

Introduce `void*` parameter for user data in listeners to let callbacks keep context on demand.

### DIFF
--- a/core/pbcc_subscribe_event_listener.c
+++ b/core/pbcc_subscribe_event_listener.c
@@ -45,6 +45,19 @@ typedef struct {
 } pbcc_object_listener_t;
 
 /**
+ * @brief Real-time status update listener definition.
+ *
+ * Object listeners entry object with information about the callback to use
+ * when status update arrives.
+ */
+typedef struct {
+    /** Real-time status update handling listener function. */
+    pubnub_subscribe_status_callback_t callback;
+    /** User provided data to be passed to the callback. */
+    void* user_data;
+} pbcc_status_listener_t;
+
+/**
  * @brief Listener definition.
  *
  * Object listeners entry object with information about the source of real-time
@@ -67,6 +80,8 @@ typedef struct {
     pubnub_subscribe_listener_type type;
     /** Real-time update handling listener function. */
     pubnub_subscribe_message_callback_t callback;
+    /** User provided data to be passed to the callback. */
+    void* user_data;
     /** Object references counter. */
     pbref_counter_t* counter;
 } pbcc_listener_t;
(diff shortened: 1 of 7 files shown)


core/pbcc_subscribe_event_listener.c
```diff
@@ -248,6 +265,17 @@ static pbarray_t* pbcc_initialize_array_(
...
+    pubnub_subscribe_status_callback_t callback,
+    void*                              user_data)
+{
+    PBCC_ALLOCATE_TYPE(listener, pbcc_status_listener_t, true, NULL);
+    listener->callback  = callback;
+    listener->user_data = user_data;
+
+    return listener;
+}
+
```
parfeon, line 278: Do we actually need this? 

Can we go in other way and do the following:
1. for `pubnub_subscribe_listener_type` add `PBSL_LISTENER_ON_STATUS`
2. Modify `pbcc_event_listener_add_status_listener` and `pbcc_event_listener_remove_status_listener`?

Also we will probably will need to make a change to: 
```c
pbcc_initialize_array_(&listener->global_status, NULL))
```
and add destructor call:
```c
pbcc_initialize_array_(&listener->global_status, pbcc_listener_free_))
```

parfeon (changes requested):
There is suggestion on how we can keep things aligned without unnecessary branching.

core/pbcc_subscribe_event_listener.c
Xavrax: Hmm, but then we're mixing two different types of API and I'm not sure about it ;o

core/pbcc_subscribe_event_listener.c
```diff
@@ -589,13 +650,27 @@ void pbcc_object_listener_free_(pbcc_object_listener_t* listener)
...
+pbcc_status_listener_t* pbcc_status_listener_alloc_(
+    pubnub_subscribe_status_callback_t callback,
+    void*                              user_data)
+{
+    PBCC_ALLOCATE_TYPE(listener, pbcc_status_listener_t, true, NULL);
+    listener->counter   = pbref_counter_alloc();
```
parfeon, line 671: I see counter, but was it used anywhere? For `pbcc_listener_t` it is used with `pbcc_add_listener_` and `pbcc_remove_listener_`.

parfeon (approved):
There is a small question.

Xavrax (commented):
@KGronek-Pubnub changes seems to be correct. 
LGTB.

PR #210, opened 2025-03-19
Add timetoken parameter for `pubnub_subscribe_v2`

parfeon:
feat(subscribe): add timetoken parameter for `pubnub_subscribe_v2`

Add the ability to set `timetoken` (for message catchup) in `pubnub_subscribe_v2_options` which is used with `pubnub_subscribe_v2`.

fix(subscription): fix memory issues with freed subscriptions

Fix issue which was source of segmentation fault when tried to access memory after it has been freed.

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -1,8 +1,15 @@
 name: c-core
 schema: 1
-version: "4.18.1"
+version: "4.19.0"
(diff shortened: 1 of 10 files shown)


parfeon:
@pubnub-release-bot release

PR #211, opened 2025-04-01
Fix objects api to accept limit 100

KGronek-Pubnub:
fix(objects): fix objects api to accept `limit` with value 100

Fix the issue that there was an assertion when using App Context functions with `limit` set to 100

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -1,8 +1,13 @@
 name: c-core
 schema: 1
-version: "4.19.0"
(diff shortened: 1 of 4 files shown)


jakub-grzesiowski (approved):
LGTM!

parfeon (approved):
LGTM!

KGronek-Pubnub:
@pubnub-release-bot release

PR #217, opened 2025-06-12
fix Address Sanitiser reports in `pubnub_set_state` function

Xavrax:
fix: Fixed wrong string copy if multiple channels(groups) are provided for `pubnub_set_state`

Fixed wrong string copy if multiple channels(groups) are provided for `pubnub_set_state`. Note that using `memcpy` instead of `strncpy` is used intentionally because of the ESP support. 

fix: Fixed crash when double comma is provided to `pubnub_set_state` as a channels(groups).
 
Fixed crash when double comma is provided to `pubnub_set_state` as a channels(groups). 

fix: Fixed wrong check if the reallocation is required in `pubnub_set_state`.

Fixed wrong check if the reallocation is required in `pubnub_set_state`.

fix: Removed additional allocation of memory for temporary data.

Removed additional allocation of memory for temporary data.

fix: Changed the amount of bytes allocated based on the provided parameters instead of hardcoded values.

Changed the amount of bytes allocated based on the provided parameters instead of hardcoded values.

### DIFF
--- a/core/pbcc_set_state.c
+++ b/core/pbcc_set_state.c
@@ -67,8 +67,6 @@ void pbcc_adjust_state(struct pbcc_context* core,
                 if (ch_temp == NULL) { end = true; ch_len = strlen(str_ch); }
                 else { ch_len = ch_temp - str_ch; }
 
-                if (ch_len == 0) { continue; }
-
                 char* curr_ch = (char*)malloc(ch_len + 1);
                 memcpy(curr_ch, str_ch, ch_len);
                 curr_ch[ch_len] = '\0';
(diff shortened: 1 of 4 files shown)


core/pbcc_set_state.c
```diff
@@ -87,29 +89,32 @@ void pbcc_adjust_state(struct pbcc_context* core,
...
-                free(curr_cg);
             } while (false == end);
         }
 
         int cb_len = strlen("}");
-        memcpy(json_state + mem_len, "}", cb_len);
-        mem_len += cb_len;
+        int trim_comma = json_state[mem_len - 1] == ',' ? 1 : 0;
+        memcpy(json_state + mem_len - trim_comma, "}", cb_len);
+        mem_len += cb_len - trim_comma;
```
parfeon, line 117: If we want to do this, should it be like this:
```suggestion
        int trim_comma = json_state[mem_len - 1] == ',' ? strlen(",") : 0;
        memcpy(json_state + mem_len - trim_comma, "}", cb_len);
        mem_len += cb_len - trim_comma;
```
?

PR #219, opened 2025-06-22
Fix issue because of which the sync operation doesn't cancel

parfeon:
fix(cancel): fix issue because of which the sync operation doesn't cancel

Fix issue because of which it was impossible for sync PubNub context version to cancel operation awaited in secondary thread,

### DIFF
--- a/core/pubnub_internal_common.h
+++ b/core/pubnub_internal_common.h
@@ -414,7 +414,10 @@ struct pubnub_ {
 
 #if PUBNUB_THREADSAFE
     pubnub_mutex_t monitor;
+    pubnub_mutex_t  cancel_monitor;
 #endif
+    /** Whether sync `await` should stop (cancel) or not. */
+    bool should_stop_await;
 
 #if PUBNUB_TIMERS_API
     /** Duration of the transaction timeout, in milliseconds */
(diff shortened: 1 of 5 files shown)


core/pubnub_internal_common.h
```diff
@@ -414,7 +414,10 @@ struct pubnub_ {
 
 #if PUBNUB_THREADSAFE
     pubnub_mutex_t monitor;
+    pubnub_mutex_t  cancel_monitor;
 #endif
+    /** Whether sync `await` should stop (cancel) or not. */
+    bool should_stop_await;
```
Xavrax, line 420: Is it guarded only for `sync` version?

core/pubnub_pubsubapi.c
```diff
@@ -81,6 +82,7 @@ pubnub_t* pubnub_init(pubnub_t* p, const char* publish_key, const char* subscrib
     p->state                          = PBS_IDLE;
     p->trans                          = PBTT_NONE;
     p->options.use_http_keep_alive    = true;
+    p->should_stop_await              = false;
```
Xavrax, line 85: Shouldn't it be only for sync version?

Xavrax (approved):

parfeon:
@pubnub-release-bot release

PR #221, opened 2025-07-11
Fix pbpal_mbedtls deinitialization/free

blake-spangenberg:
For the mbedTLS implementation:
- Fixes memory leak when calling `pubnub_free` on an initialized PubNub object
- Fixes socket leak when calling `pubnub_free` on a PubNub object with an open connection
- Fixes socket leak when publishing using a context with `use_http_keep_alive = 0`
- Fixes crash due to double-free when calling `pubnub_free` after `pbpal_handle_socket_condition` encounters a socket error


Xavrax:
Hello @blake-spangenberg !
Thank you for your contribution. 
Because of our internal policies I had reopen the PR to launch the GH checks ( #223 ). 
I've added your credits to CHANGELOG. 

Thank you so much!

PR #227, opened 2025-07-28
Guard `unistd` inclusion with platform check

parfeon:
fix(samples): define sleep function for Windows builds

refactor(include): guard `unistd` inclusion with platform check


Xavrax (approved):
LGTM

PR #231, opened 2025-09-22
Address issue that returned `poll` immediately

parfeon:
fix(poller): address issue that returned `poll` immediately

Fix the issue because of which `poll` always returned immediately without waiting for the actual event (TLS handshake flow uses proper watcher now).

### DIFF
--- a/core/pbpal.h
+++ b/core/pbpal.h
@@ -37,13 +37,15 @@ enum pbpal_resolv_n_connect_result {
 enum pbpal_tls_result {
     pbtlsEstablished,
     pbtlsStarted,
+    pbtlsStartedWaitRead,
+    pbtlsStartedWaitWrite,
     pbtlsInProgress,
     pbtlsResourceFailure,
     pbtlsFailed
 };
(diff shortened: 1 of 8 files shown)


core/pbpal.h
```diff
@@ -37,13 +37,15 @@ enum pbpal_resolv_n_connect_result {
...
+    pbtlsStartedWaitRead,
+    pbtlsStartedWaitWrite,
     pbtlsInProgress,
     pbtlsResourceFailure,
     pbtlsFailed
 };
 
 /* Handles socket condition on given platform */
-enum pubnub_res pbpal_handle_socket_condition(int result, pubnub_t* pb, char const* file, int line);
+enum pubnub_res pbpal_handle_socket_condition(int result, pubnub_t* pb, char const* file, int line, bool *needRead, bool* needWrite);
```
Xavrax, line 48: I would recommend using tribool instead of bool ptr.

Xavrax (approved):
LGTM

PR #232, opened 2025-10-20
Add ability to configure TCP keepalive

parfeon:
feat(tcp-keepalive): add ability to configure TCP keepalive

Provide interfaces (`pubnub_use_tcp_keep_alive`, `pubnub_dont_use_tcp_keep_alive`, `use_tcp_keep_alive`, and `dont_use_tcp_keep_alive`) to set up active TCP connection keep-alive packet sending.

fix(dns): fix crash on Windows system DNS
    
Replace the Windows API, which in a multithreaded environment became a reason for crashes during the DNS resolution process.

fix(sockets): address issue with additional socket

Fix the issue because of which client created a secondary connection when built with `PUBNUB_USE_IPV6` support.

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -1,8 +1,17 @@
 name: c-core
 schema: 1
-version: "5.3.2"
+version: "5.4.0"
 scm: github.com/pubnub/c-core
 changelog:
+  - date: 2025-10-21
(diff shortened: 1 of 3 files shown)


parfeon:
@pubnub-release-bot release

PR #233, opened 2025-11-05
Add limit and offset parameters to pubnub_here_now functionality.

KGronek-Pubnub:
feat(here_now): Add `limit` and `offset` parameters to `pubnub_here_now` functionality.

Add `limit` and `offset` parameters to `pubnub_here_now_options` used in `pubnub_here_now_ex`. Set `limit` to be set to default (`1000`) for `pubnub_here_now` (not ex) function.

BREAKING CHANGES: `here_now` by default will return `1000` users at once.

refactor(dns-server): use `DnsQueryConfig`

Try receiving DNS servers with `DnsQueryConfig` and fallback to the adapters with `GetAdaptersAddresses`.

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -1,8 +1,15 @@
 name: c-core
 schema: 1
-version: "5.4.0"
+version: "6.0.0"
 scm: github.com/pubnub/c-core
 changelog:
+  - date: 2025-11-07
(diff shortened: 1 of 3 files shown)


parfeon (approved):
But in general LGTM!

PR #235, opened 2025-11-12
Add SNI/hostname verification support for TLS connections and DNS server discovery improvement

parfeon:
feat(ssl): add SNI support for TLS connections
    
Added SNI (Server Name Indication) support for TLS connections.
    
feat(ssl): add hostname verification
    
Added hostname verification for SSL/TLS certificates to prevent MITM attacks.

refactor(ssl): actualize bundled certificates

Update bundled certificates (added Amazon Root CA 1) which is used for secured connection and peers verification.

refactor(dns-windows): improve DNS server discovery

Filter out DNS servers received from disconnected adapters, loopback, and APIPA addresses.

test(dns-tls-windows): add test to test DNS discovery

Added tests to test DNS server discovery in a controlled Windows environment (manual).

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -7,9 +7,9 @@ changelog:
     version: v6.1.0
     changes:
       - type: feature
-        text: "Add SNI support for TLS connections."
+        text: "Added SNI (Server Name Indication) support for TLS connections."
       - type: feature
-        text: "Added SNI (Server Name Indication) support for TLS connections.           Added hostname verification for SSL/TLS certificates to prevent MITM attacks."
+        text: "Added hostname verification for SSL/TLS certificates to prevent MITM attacks."
       - type: improvement
         text: "Update bundled certificates (added Amazon Root CA 1) which is used for secured connection and peers verification."
       - type: improvement
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 November 17 2025
 
 #### Added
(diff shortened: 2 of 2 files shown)


windows/pubnub_dns_system_servers.c
```diff
@@ -8,17 +8,156 @@
...
+ *  1. Adapters on the best route (lowest priority number)
+ *  2. Lower interface metric
+ *
+ *  @param a Left-hand @c adapter_with_metric which will be used in comparison.
+ *  @param b Right-hand @c adapter_with_metric which will be used in comparison.
+ *  @return @b -1 when the left-hand adapter has higher priority than the
+ *  right-hand; @b 0 when both adapters have the same priority; @b 1 when the
+ *  right-hand adapter has higher priority than the left-hand.
+ */
+static int compare_adapters_by_priority(const void* a, const void* b)
```
Xavrax: why `void*`?

windows/tests/pubnub_dns_windows_test.cpp
```diff
@@ -0,0 +1,999 @@
+/* -*- c-file-style:"stroustrup"; indent-tabs-mode: nil -*- */
```
Xavrax: Why do we test like this? 
Have we changed the testing approach?

Xavrax: Why we test like this?

PR #238, opened 2026-02-19
Advanced logging and subscription / subscription set release fix

parfeon:
feat(logger): added new advanced logger

Added support for extendable logging, which can log structured data and allow to register customer loggers.

BREAKING CHANGES: Changes in log macros, functions, and custom log callback removal will require code changes.

feat(logger): added default logger

Added default logger for POSIX/Windows (`stdio`), and separate for FreeRTOS and Harmony.

feat(log): added request / response logging

Now request / response is logged as single log entries.

feat(log): rapid `TRACE` logs suppression

Added mechanism to drastically reduce the amount of logs generated with `sync` await interface.

feat(pn_sdk): added way to set pn_sdk in runtime

fix(subscription): fix pointer nullify on unsubscribe

Fix issue because of which `pubnub_unsubscribe_with_subscription` and `pubnub_unsubscribe_with_subscription_set` nullified passed pointer even if 'user' code still owns it.

fix(subscription): fix use-after-free on mutex

Save mutex to local variable before freeing subscription and subscription_set to avoid dereferencing freed memory on unlock/destroy.

fix(subscribe-ee): unregister callback before freeing event engine

Unregister PubNub callback in `pbcc_subscribe_ee_free()` so late async completions do not invoke it with freed `ee` as user_data.

fix(event-listener): remove status listener by callback and user_data

`pbarray_remove()` uses pointer equality; match by (callback, user_data) and remove by index with `pbarray_remove_element_at()` so the intended listener is actually removed.

fix(crypto): fix null pointer dereference in cryptor header

Move NULL check for `cryptor_header` before the first dereference in `pbcc_cryptor_header_v1_to_alloc_block` and remove the redundant check.

fix(subscription): fix null pointer dereference in subscription set

Guard `subs` dereference in `pubnub_subscription_set_union` and `pubnub_subscription_set_subtract` with a NULL check before the loop.

fix(actions): fix adding long message actions by disabling Gzip compression

fix(dns): fix type mismatch in POSIX system DNS server reader

Access `.ipv4` member when assigning parsed address to `dns_bytes`.

fix(ipv6): detect false-positive IPv6 support on Windows

Use `getsockname()` to verify a Global Unicast source (2000::/3) instead of trusting `connect()`.

refactor(tls): fix TLS handshake busy-wait in sync mode

Fixed issues because of which TLS read returned right away to the state machine without even waiting for `select`.

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -17,7 +17,7 @@ changelog:
       - type: feature
         text: "Added way to set pn_sdk in runtime."
       - type: bug
-        text: "Fix issue because of which `pubnub_unsubscribe_with_subscription` and `pubnub_unsubscribe_with_subscription_set` nullified passed pointer even if "user" code still owns it."
+        text: "Fix issue because of which `pubnub_unsubscribe_with_subscription` and `pubnub_unsubscribe_with_subscription_set` nullified passed pointer even if 'user' code still owns it."
       - type: bug
         text: "Save mutex to local variable before freeing subscription and subscription_set to avoid dereferencing freed memory on unlock/destroy."
       - type: bug


core/pubnub_subscribe_event_engine.c
```diff
@@ -539,21 +511,17 @@ enum pubnub_res pubnub_unsubscribe_with_subscription_set(
...
-    if (PNR_OK == rslt || PNR_SUB_NOT_FOUND == rslt) {
-        /**
-         * Decrease subscription set reference counter only if it has been used
-         * for previous subscriptions.
-         */
-        if (PNR_OK == rslt)
-            subscription_set_reference_count_update_(*set, false);
+    if (ref_counter > 1) *set = set_ref;
 
+    if (PNR_OK == rslt || PNR_SUB_NOT_FOUND != rslt) {
```
KGronek-Pubnub: Does this condition make sense? Shouldn't it be `(PNR_OK == rslt || PNR_SUB_NOT_FOUND == rslt)`?

KGronek-Pubnub:
Issue found by AI:

This is a real UAF risk:
these statements coexist in `core/pubnub_subscribe_event_engine.c`:

- pubnub_subscription_free():

```
locks (*sub)->mutw
calls pubnub_subscription_free_(*sub)
then does pubnub_mutex_unlock((*sub)->mutw);
then pubnub_mutex_destroy((*sub)->mutw);
```

- pubnub_subscription_free_():

`when pbref_counter_free(sub->counter) == 0, it calls free(sub);`
So the problematic sequence is:

1. pubnub_subscription_free() holds (*sub)->mutw.
2. pubnub_subscription_free_() frees sub.
3. Control returns to pubnub_subscription_free().
4. It dereferences (*sub) (freed memory) to unlock/destroy mutex.

That dereference after free(sub) is undefined behavior (classic UAF read/use).

core/pubnub_dns_servers.c
KGronek-Pubnub: Build error:
```suggestion
    ret = pubnub_parse_ipv6_addr(NULL, ipv6_str, &m_secondary_dns_server_ipv6);
```

or 

```suggestion
    ret = pubnub_parse_ipv6_addr(pb, ipv6_str, &m_secondary_dns_server_ipv6);
```

core/pubnub_ccore_pubsub.c
```diff
@@ -54,10 +59,26 @@ void pbcc_init(struct pbcc_context* p,
...
+    if (0 != pubnub_generate_uuid_v4_random(&uuid)) bzero(p->id, sizeof(p->id));
+    else {
+        const struct Pubnub_UUID_String str_uuid = pubnub_uuid_to_string(&uuid);
+        memcpy(p->id, str_uuid.uuid, sizeof(p->id));
+    }
+#if PUBNUB_USE_LOGGER
+    p->logger_manager = pbcc_logger_manager_alloc(p->id);
+#if PUBNUB_USE_DEFAULT_LOGGER
+    pbcc_logger_manager_set_default_logger(
+        p->logger_manager, pubnub_default_logger_alloc());
```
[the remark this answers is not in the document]
KGronek-Pubnub: Sure I get the point... well, I don't know what is better, crash or some undefined behavior :) so I leave it up to you to decide.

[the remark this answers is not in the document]
parfeon: I made changes here as well.

core/pubnub_subscribe_event_engine.c
parfeon: @KGronek-Pubnub I've simplified condition. The rest of the notes also has been addressed.

KGronek-Pubnub (approved):
Verified on Windows - compiles and works great. Good job, LGMT :)

PR #244, opened 2026-03-11
Fix subscribe crash (SIGABRT) on Android/OpenSSL.

KGronek-Pubnub:
fix(include): fix subscribe SIGABRT on OpenSSL by using include path for` pubnub_internal.h`

Fix ABI mismatch where posix/windows sources picked the local `pubnub_internal.h` (wrong struct pubnub_pal size) on OpenSSL builds, so timeout fields were read as 0 and triggered an assertion. Use #include <pubnub_internal.h> so the include path selects the correct header per build.

fix(version): add missing pubnub_uagent for OpenSSL platform

Add missing `pubnub_uagent()` function implementation for OpenSSL platform as declared in `pubnub_version.h`. Update `pubnub_uname()` format to match Windows and POSIX implementations and add `PUBNUB_SDK_VERSION_SUFFIX` support.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -21,8 +21,8 @@ jobs:
         runner:
           - os: ubuntu-latest
             group: organization/Default
-#          - os: macos-13
-#            group: organization/macos-gh
+    #          - os: macos-13
+    #            group: organization/macos-gh
     steps:
       - name: Checkout project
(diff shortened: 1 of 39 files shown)


parfeon (approved):
LGTM!

KGronek-Pubnub:
@pubnub-release-bot release

PR #245, opened 2026-03-12
DNS txnid and static analysis

### LINKED ISSUE #243: DNS request abusing the RFC by reusing the Request-Id
We have recently upgraded our `pubnub:c-core` from [v5.1.0](https://github.com/pubnub/c-core/releases/tag/v5.1.0) to [v7.0.2](https://github.com/pubnub/c-core/releases/tag/v7.0.2).

**due to this upgrade our pubunb integration stopped working.**

## issue description
after deep analysis we found out that:
- earlier version was sending a single DNS request for the pubnub FQDN either `A` record in case ipv4(`pubnub_set_ipv4_connectivity`) was set or `AAAA` if ipv6 (`pubnub_set_ipv6_connectivity`) was set.
- the newer version sends 2 DNS requests `A` & `AAAA` regardless the type of network being set with the APIs mentioned above.

## easy solution
we have made a change to fix this and we can suggest it uplink.
- as your code is not checking the DNS response TxnID (which again is not according to RFC) the easy solution would be to pick a random request-id.
- a whole solution would be to check keep the request-id as a reference and check that the response is actually related, but we understand that this would require more time.

### suggested code fix
```c++
diff --git a/lib/pubnub_dns_codec.c b/lib/pubnub_dns_codec.c
index 31c8e7a..fde580a 100644
--- a/lib/pubnub_dns_codec.c
+++ b/lib/pubnub_dns_codec.c
@@ -158,15 +158,9 @@ int pbdns_prepare_dns_request(
     int*              to_send,
     enum DNSqueryType query_type)
 {
-    static bool srand_called = false;
     int qname_encoded_length;
     int len = 0;
 
-    if (!srand_called) {
-        srand_called = true;
-        srand(time(NULL));
-    }
-
     PUBNUB_ASSERT_OPT(buf != NULL);
     PUBNUB_ASSERT_OPT(host != NULL);
     PUBNUB_ASSERT_OPT(to_send != NULL);
@@ -183,7 +177,7 @@ int pbdns_prepare_dns_request(
         return -1;
     }
     buf[HEADER_ID_OFFSET]              = 0;
-    buf[HEADER_ID_OFFSET + 1]          = rand() % 0xFFFF;
+    buf[HEADER_ID_OFFSET + 1]          = 33; /* in lack of a better ID */
     buf[HEADER_OPTIONS_OFFSET]         = dnsoptRDmask >> 8;
     buf[HEADER_OPTIONS_OFFSET + 1]     = 0;
     buf[HEADER_QUERY_COUNT_OFFSET]     = 0;

```

parfeon:
fix(url): fix dangling pointer in `here_now` params

Move `offset_buf` declaration outside the `if` block so it remains in scope when `ENCODE_URL_PARAMETERS` reads the stored pointer.

fix(crypto): fix `free()` on string literal

Initialize `part_sign` to `NULL` instead of `(char*)""` to avoid undefined behavior when `free()` is called in non-crypto builds.

fix(dns): generate unique transaction IDs per query

Each DNS query now gets a random 16-bit transaction ID via `rand()`, seeded once with `time(NULL)`. IDs are stored in `dns_queries_tracking` and validated on response to ensure concurrent A and AAAA queries have distinct IDs per RFC 5452.

Closes #243

fix(dns): filter multicast, reserved and broadcast IPv4 addresses

Filter DNS server addresses with first `octet >= 224` in `is_valid_ipv4`. Multicast, reserved/Class E, and broadcast ranges are not valid DNS server addresses.

fix: resolve AB-BA deadlock between `ee->mutw` and `pb->monitor` in subscribe EE

Release `ee->mutw` before calling PubNub APIs or `pbcc_ee_handle_event` to prevent lock-order inversion with the IO callback thread in both subscribe and unsubscribe paths. Add deadlock regression test covering all unsubscribe paths.

fix: defer `m_lock` in `pballoc_free_at_last`

Move `m_lock` acquisition after `pbcc_deinit()`/`pbpal_free()` because they don't need the global allocator lock and `pbcc_subscribe_ee_free` re-locks the non-recursive `m_lock` via `pubnub_register_callback` → `pb_valid_ctx_ptr`, causing self-deadlock on the same thread.

fix: remove dangling `pb` dereference in `pubnub_entity_free`

Entity can outlive its parent `pubnub_t` context, so logging through `_entity->pb` is a use-after-free.

fix: upgrade DNS validation probe to standard A-query

Replace header-only probe (`QDCOUNT=0`) with a proper A-query for `example.com` and validate response transaction ID, fixing false negatives on resolvers that silently drop malformed packets.

build(log): fix `winsock2.h` include order

Move `pubnub_internal.h` include above `windows.h` in `pbcc_logger_manager.c` so `winsock2.h` is included first, preventing redefinition errors in builds without `_WINSOCKAPI_` defined.

refactor(url): fix format specifier for unsigned count

Use `%u` instead of `%d` for `unsigned int` `count` in `snprintf` to match the actual type.

test(dns): expand Windows DNS discovery test suite

Add buffer edge cases, result stability, broadcast/multicast filtering, no-DNS adapter handling, adapter flapping stress test, and multiple DNS per adapter phases. Extend baseline with multicast check and IPv6 with duplicate detection.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -66,6 +66,8 @@ jobs:
           ref: v1
           token: ${{ secrets.GH_TOKEN }}
           path: .github/.release/actions
+      - name: Install devcon
+        run: choco install devcon.portable -y
       - name: Setup MSVC environment
         uses: ilammy/msvc-dev-cmd@v1
       - name: Build DNS discovery test
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -99,10 +99,42 @@ jobs:
             windows\pubnub_dns_system_servers.c ^
             core\pubnub_assert_std.c ^
             ws2_32.lib iphlpapi.lib
+      - name: Build DNS discovery test (with validation)
+        shell: cmd
+        run: |
+          cl /nologo /W3 /Fe:tests\dns_discovery\dns_discovery_test_validated.exe ^
+            /DPUBNUB_CALLBACK_API ^
+            /DPUBNUB_USE_IPV6=1 ^
+            /DPUBNUB_USE_SSL=0 ^
+            /DPUBNUB_SET_DNS_SERVERS=1 ^
+            /DPUBNUB_CHANGE_DNS_SERVERS=0 ^
+            /DPUBNUB_USE_LOGGER=1 ^
+            /DPUBNUB_USE_DEFAULT_LOGGER=0 ^
+            /DPUBNUB_LOG_MIN_LEVEL=NONE ^
+            /DPUBNUB_DYNAMIC_REPLY_BUFFER=1 ^
+            /DPUBNUB_RECEIVE_GZIP_RESPONSE=0 ^
+            /DPUBNUB_ADVANCED_KEEP_ALIVE=0 ^
+            /DPUBNUB_USE_SUBSCRIBE_EVENT_ENGINE=0 ^
+            /DPUBNUB_USE_SUBSCRIBE_V2=0 ^
+            /DPUBNUB_CRYPTO_API=0 ^
+            /DPUBNUB_USE_GZIP_COMPRESSION=0 ^
+            /DPUBNUB_USE_GRANT_TOKEN_API=0 ^
+            /DPUBNUB_USE_REVOKE_TOKEN_API=0 ^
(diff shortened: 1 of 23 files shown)


KGronek-Pubnub (approved):
LGTM

parfeon:
@pubnub-release-bot release

PR #246, opened 2026-04-02
feat/ Allow users to select if they want to cache or consume the value of the cypher key depending on their crypto module implementation

Xavrax:
feat: Allow users to select if they want to cache or consume the value of the cypher key depending on their crypto module implementation

Allow users to select if they want to cache or consume the value of the cypher key depending on their crypto module implementation

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -1,8 +1,13 @@
 name: c-core
 schema: 1
-version: "7.1.3"
+version: "7.2.0"
 scm: github.com/pubnub/c-core
 changelog:
+  - date: 2026-04-22
+    version: v7.2.0
+    changes:
+      - type: feature
+        text: "Allow users to select if they want to cache or consume the value of the cypher key depending on their crypto module implementation."
   - date: 2026-04-16
     version: v7.1.3
     changes:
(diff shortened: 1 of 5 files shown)


KGronek-Pubnub:
I'm not sure is it still relevant for us to support C++ 98? Some concepts like `std::move`, are C++ 11, and they are in main wrapper file `pubnub_common.hpp`, so I guess it would mean - end of C++ 98 support? Maybe we should plan to update some comments/flags accordingly if so.

Xavrax:
> I'm not sure is it still relevant for us to support C++ 98? Some concepts like `std::move`, are C++ 11, and they are in main wrapper file `pubnub_common.hpp`, so I guess it would mean - end of C++ 98 support? Maybe we should plan to update some comments/flags accordingly if so.

Good catch. I forgot about the compatibility of other versions but the Crypto module itself is no longer C++98 compatible. Should I make it compatible again?
@KGronek-Pubnub 
@parfeon

parfeon:
@Xavrax Do we actually need C++98 support? Event C99 currently exists mostly to support older toolchains for IoT devices but they already don't need it and can normally operate with latest standards.
@KGronek-Pubnub what do you think?

KGronek-Pubnub:
@Xavrax @parfeon  I don't see any real problem with skipping C++ 98 support, just wanted to mention that if we do it, let's do it intentionally :)

KGronek-Pubnub (approved):
LGTM!

Xavrax:
@pubnub-release-bot release

PR #248, opened 2026-04-24
Fix/cancelation

Xavrax:
fix: Initialise logger request/response buffers in pbcc_init to make sure that the loggers doesn't receive junked data when cancelation occurs before request preparation. 

Initialise logger request/response buffers in pbcc_init to make sure that the loggers doesn't receive junked data when cancelation occurs before request preparation. 

fix: Fixed data race in situation when one logger used via logger_manager is being called from two different contexts.

Fixed data race in situation when one logger used via logger_manager is being called from two different contexts.

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -1,8 +1,15 @@
 name: c-core
 schema: 1
-version: "7.2.0"
+version: "7.2.1"
 scm: github.com/pubnub/c-core
 changelog:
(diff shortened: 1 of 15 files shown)


KGronek-Pubnub (approved):
LGTM

parfeon:
@pubnub-release-bot release

PR #249, opened 2026-05-13
Replace `select()` with `poll()`

parfeon:
fix(socket): replace `select()` with `poll()` in `pbpal_check_connect`

Replace inline `select()`/`FD_SET` with `poll()` on non-Windows platforms in `pbpal_check_connect()`. On Linux and macOS `FD_SET` writes to a fixed 1024-bit bitmap and corrupts the stack when a socket fd value exceeds `FD_SETSIZE`. `poll()` has no fd-value limitation. Windows retains `select()` because its `fd_set` is array-based and safe.

test(poller): add unit tests for `poll()`-based callback poller

Add `pbpal_ntf_callback_poller_poll_unit_test.c` covering init/deinit, save/remove, watch events, poll detection of readable and writable sockets, multiple sockets, socket update, and a high fd number test case (fd=1500) that validates the fix for the `FD_SETSIZE` overflow.

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -1,8 +1,15 @@
 name: c-core
 schema: 1
-version: "7.2.1"
+version: "7.2.2"
 scm: github.com/pubnub/c-core
 changelog:
+  - date: 2026-05-14
+    version: v7.2.2
+    changes:
+      - type: bug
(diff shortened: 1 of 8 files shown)


KGronek-Pubnub (approved):
LGTM

parfeon:
@pubnub-release-bot release

PR #250, opened 2026-05-20
Fix subscribables options override and event listener refcount.

KGronek-Pubnub:
fix(subscription): Caller options now override presence flag; no silent ignore.

When building a subscription’s subscribable set, caller-supplied options now actually override `receive_presence_events` instead of copying `sub->options` again, so set-level presence settings are applied consistently.

fix(listener): One refcount drop per remove; transient add ref released—no UAF or leak.

Removal no longer decrements the listener refcount twice (once manually and once in the array destructor), and add paths release the allocator's transient ref at the end, so shared listeners are not freed too early (use-after-free for N≥3 arrays) and global listeners are not leaked after a single remove.

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -1,8 +1,15 @@
 name: c-core
 schema: 1
-version: "7.2.2"
+version: "7.2.3"
 scm: github.com/pubnub/c-core
 changelog:
+  - date: 2026-05-21
+    version: v7.2.3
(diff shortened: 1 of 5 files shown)


parfeon (approved):
LGTM!

KGronek-Pubnub:
@pubnub-release-bot release

PR #251, opened 2026-06-18
Fix a callback API crash on Windows during shutdown/reconnect.

KGronek-Pubnub:
fix(crash): Fix timer-list assert in callback API on Windows

Fix timer-list assert in callback API on Windows when closing or reconnecting during async DnsQueryEx DNS resolution.

refactor(callback): Harden callback API timer-list handling on all platforms

Harden callback API timer-list handling on all platforms: remove before re-add in `pbntf_got_socket_callback()`, and always unregister watcher/timer/queue state from `pbpal_close()/pbpal_free()` on callback builds.

### DIFF
--- a/lib/sockets/pbpal_sockets.c
+++ b/lib/sockets/pbpal_sockets.c
@@ -5,9 +5,6 @@
 #include "core/pubnub_ntf_sync.h"
 #include "core/pubnub_netcore.h"
 #include "core/pubnub_assert.h"
-#if defined(_WIN32) && defined(PUBNUB_CALLBACK_API)
-#include "windows/pbpal_dns_query_ex.h"
-#endif
 #if PUBNUB_USE_LOGGER
 #include "core/pubnub_logger.h"
 #endif // PUBNUB_USE_LOGGER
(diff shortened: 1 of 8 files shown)


posix/pubnub_ntf_callback_posix.c
```diff
@@ -259,6 +259,7 @@ MAYBE_INLINE int pbntf_got_socket_callback(pubnub_t* pb)
 
     if (PUBNUB_TIMERS_API) {
         pthread_mutex_lock(&m_watcher.timerlock);
+        pbpal_remove_timer_safe(pb, &m_watcher.timer_head);
```
Xavrax, line 262: The change is affecting not only the Windows ecosystem.

lib/sockets/pbpal_sockets.c
```diff
@@ -305,14 +302,13 @@ void pbpal_forget(pubnub_t* pb)
...
-#if defined(_WIN32) && defined(PUBNUB_CALLBACK_API)
-    /* Cancel any pending DnsQueryEx queries. Must happen regardless of
-       socket state because the DnsQueryEx path does not create a UDP
-       socket (pb->pal.socket stays SOCKET_INVALID). */
-    pbpal_os_dns_cancel(pb);
+#if defined(PUBNUB_CALLBACK_API)
+    /* Unregister from watcher/timer/queue even when no socket exists yet
+       (e.g. Windows DnsQueryEx async DNS). pbntf_lost_socket() also
+       cancels pending DnsQueryEx on Windows. */
+    pbntf_lost_socket(pb);
```
Xavrax, line 309: So my inderstanding is that the pbntf_lost_socket calls the pbpal_os_dns_cancel somewhere in the code, is my assumption correct?

Xavrax (approved):
LGTM (anwer before merge)

posix/pubnub_ntf_callback_posix.c
KGronek-Pubnub: The crash was most likely happening on Windows, but defensive hardening was also implemented to Posix.

lib/sockets/pbpal_sockets.c
KGronek-Pubnub: Yes on Windows callback.
```
MAYBE_INLINE void pbntf_lost_socket_callback(pubnub_t* pb)
{
    pbpal_os_dns_cancel(pb);
    EnterCriticalSection(&m_watcher.mutw);
    pbpal_ntf_callback_remove_socket(m_watcher.poll, pb);
    ...
    pbpal_remove_timer_safe(pb, &m_watcher.timer_head);
    ...
```

posix/pubnub_ntf_callback_posix.c
Xavrax: so it should be added to the change log. It can (probably not but still) fail something on posix side, so it would be nice to see that info in the change log.

KGronek-Pubnub:
@pubnub-release-bot release

PR #252, opened 2026-07-27
Reset `http_buf_len` on NTLM proxy retry to prevent assert crash

parfeon:
feat(ntlm): parse domain from username in SSPI identity for NTLM proxy auth

`fill_sspi_identity` now splits `DOMAIN\user` and `user@domain` formats into separate User and
Domain fields so proxies requiring domain-qualified credentials can authenticate.

fix(proxy): reset `http_buf_len` on NTLM proxy retry to prevent assert crash

Reset `http_buf_len` to 0 in `pbproxyFinRetry` so the stale 407 body length does not trip the
assert guarding `proxy_saved_path` memcpy on the next CONNECT attempt.

fix(ntlm): correct inverted SSPI max-token guard in `pbntlm_packer_sspi.c`

The warning fired on the normal case instead of when SSPI needs more space than our buffer provides.
Flip the comparison and update the log message.

refactor(ntlm): use dynamic allocation for Type-2 challenge decode in `pbntlm_core_handle`

Replace the 512-byte stack buffer with `pbbase64_decode_alloc_std` so large NTLM Type-2 challenges
from domain-joined proxies are no longer silently dropped.

refactor(ntlm): raise `PUBNUB_NTLM_MAX_TOKEN` from 1024 to 4096

NTLMv2 Type-3 with extended target info and MIC frequently exceeds 1024 bytes. Raised to 4096 to
cover all known SSPI scenarios.

test(proxy): add NTLM proxy regression tests

Add tests for oversized 407 body triggering retry-path assert, and direct `pbntlm_core_handle`
tests with 600, 2048, and 4200 byte challenges.

### DIFF
--- a/core/pbntlm_core.c
+++ b/core/pbntlm_core.c
@@ -8,6 +8,8 @@
 
 #include "lib/base64/pbbase64.h"
 
+#include <stdlib.h>
+
 
 void pbntlm_core_init(pubnub_t* pb)
 {
@@ -25,9 +27,7 @@ void pbntlm_core_deinit(pubnub_t* pb)
 
 void pbntlm_core_handle(pubnub_t* pb, char const* base64_msg, size_t length)
 {
-    int             i;
-    uint8_t         msg[512];
-    pubnub_bymebl_t data = { msg, sizeof msg / sizeof msg[0] };
+    pubnub_bymebl_t data;
 
     if (pbntlmDone == pb->ntlm_context.state) {
         pbntlm_core_init(pb);
@@ -42,14 +42,15 @@ void pbntlm_core_handle(pubnub_t* pb, char const* base64_msg, size_t length)
         pbntlm_core_deinit(pb);
         return;
     }
-    i = pbbase64_decode_std(base64_msg, length, &data);
-    if (0 != i) {
-        PUBNUB_LOG_ERROR(pb, "Base64 decode failed with error code: %d", i);
+    data = pbbase64_decode_alloc_std(base64_msg, length);
+    if (NULL == data.ptr) {
+        PUBNUB_LOG_ERROR(pb, "Base64 decode (alloc) of NTLM challenge failed");
         pbntlm_core_deinit(pb);
         return;
     }
     (void)pbntlm_unpack_type2(pb, &pb->ntlm_context, data);
     pb->ntlm_context.state = pbntlmSendAuthenticate;
+    free(data.ptr);
 }
 
 
--- a/core/pbntlm_core.c
+++ b/core/pbntlm_core.c
@@ -48,7 +48,12 @@ void pbntlm_core_handle(pubnub_t* pb, char const* base64_msg, size_t length)
         pbntlm_core_deinit(pb);
         return;
     }
-    (void)pbntlm_unpack_type2(pb, &pb->ntlm_context, data);
+    if (0 != pbntlm_unpack_type2(pb, &pb->ntlm_context, data)) {
+        PUBNUB_LOG_ERROR(pb, "Failed to unpack NTLM Type-2 challenge");
+        free(data.ptr);
+        pbntlm_core_deinit(pb);
+        return;
+    }
     pb->ntlm_context.state = pbntlmSendAuthenticate;
     free(data.ptr);
 }
--- a/core/pbntlm_packer_sspi.c
+++ b/core/pbntlm_packer_sspi.c
@@ -6,6 +6,8 @@
 
 #include "pubnub_assert.h"
 
+#include <string.h>
+
 
 #pragma comment(lib, "secur32")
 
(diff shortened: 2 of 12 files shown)


core/pbntlm_core.c
```diff
@@ -42,14 +42,15 @@ void pbntlm_core_handle(pubnub_t* pb, char const* base64_msg, size_t length)
         pbntlm_core_deinit(pb);
         return;
     }
-    i = pbbase64_decode_std(base64_msg, length, &data);
-    if (0 != i) {
-        PUBNUB_LOG_ERROR(pb, "Base64 decode failed with error code: %d", i);
+    data = pbbase64_decode_alloc_std(base64_msg, length);
+    if (NULL == data.ptr) {
+        PUBNUB_LOG_ERROR(pb, "Base64 decode (alloc) of NTLM challenge failed");
```
Xavrax, line 47: I like these change in case of readability, however we're losing potentially part of the info (I see that it might return values like `-14` and `-15`. (which is meaningless without a code to debug, but still).

core/pbntlm_packer_sspi.c
```diff
@@ -52,11 +74,13 @@ void pbntlm_packer_init(pubnub_t* context, pbntlm_ctx_t* pb)
...
+    if (sspi_max_token > PUBNUB_NTLM_MAX_TOKEN) {
         PUBNUB_LOG_WARNING(
             context,
-            "NTLM token for SSPI is smaller (%lu) that maximum allowed (%d).",
+            "SSPI max token (%lu) exceeds PUBNUB_NTLM_MAX_TOKEN (%d). "
+            "Type-2 challenges larger than %d bytes will be rejected.",
             sspi_max_token,
+            PUBNUB_NTLM_MAX_TOKEN,
             PUBNUB_NTLM_MAX_TOKEN);
     }
```
Xavrax, line 85: That's kinda specific change. Why we reverse that? (I mean, why the previous log is no longer needed)

core/pbntlm_core.c
```diff
@@ -42,14 +42,15 @@ void pbntlm_core_handle(pubnub_t* pb, char const* base64_msg, size_t length)
...
         pbntlm_core_deinit(pb);
         return;
     }
     (void)pbntlm_unpack_type2(pb, &pb->ntlm_context, data);
```
Xavrax: Are we sure that we can omit this value? 
As I understood, it might fail for `SPI` build (passes the `std` one).

core/pubnub_internal_common.h
```diff
@@ -160,7 +160,7 @@ enum NPBNTLM_State {
 };
 
 /** Maximum supported length of the NTLM token (message) */
-#define PUBNUB_NTLM_MAX_TOKEN 1024
+#define PUBNUB_NTLM_MAX_TOKEN 4096
```
Xavrax, line 163: We extended the maximum amount of characters for the proxy however at `core/pubnub_netcore.c:1340`, we have defined `header` variable that is hardcoded to 1024.

Xavrax (approved):
Overall looks good to me. 
Verify my comments and can be released. 

BTW. I see changes in makefiles, shouldn't we update CMake too?

core/pbntlm_core.c
Xavrax: As when it fails on SPI, there is assertion in unpack_type3 which verifies if the size of data is higher than 0 (whereas fail of this function does it). 

Ofc, handling error earlier won't safe the whole thing, but it will be consistent between build types.

core/pbntlm_packer_sspi.c
parfeon: Well, before there were strange warning, that token surprisingly smaller than configured max limit - there is nothing wrong with token to be smaller than allocated maximum size :) 
It was just misleading.

core/pbntlm_core.c
parfeon: Sadly, we will loose those but I'm not really sure about possibility that it will happen here.

core/pubnub_internal_common.h
parfeon: You're right.

Added `PUBNUB_PROXY_AUTH_HEADER_MAX_SIZE` macro that derives the proper buffer size from `PUBNUB_NTLM_MAX_TOKEN` (base64 expansion + header prefix room) and used it for `hedr` in `pubnub_netcore.c`.

core/pbntlm_core.c
parfeon: Good catch!

Added error handling - now we check the return value and deinit/abort on failure, consistent with the other error paths in this function. Also added a test for invalid base64 input.

parfeon:
> BTW. I see changes in makefiles, shouldn't we update CMake too?

Those are test apps for manual NTLM testing

parfeon:
@pubnub-release-bot release

PR #253, opened 2026-07-27
Fix infinite DNS retry loop when IPv6 DNS servers are unreachable

parfeon:
fix(dns): Fix infinite DNS retry loop when IPv6 DNS servers are unreachable

Clear the IPv6 address buffer in `get_dns_ip()` when a server is skipped due to prior failure marking, so `user_provided_ipv6_dns` correctly becomes false and the rotation falls through to IPv4 DNS servers.

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -1,8 +1,13 @@
 name: c-core
 schema: 1
-version: "7.2.4"
+version: "7.2.5"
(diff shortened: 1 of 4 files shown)


Xavrax (approved):
LGTM

parfeon:
@pubnub-release-bot release

PR #254, opened 2026-08-04
Add missing `uuid` query parameter to fetch history request

parfeon:
fix(history): add missing `uuid` query parameter to fetch history request

`pbcc_fetch_history_prep` was the only prep function not appending the `uuid` query parameter that identifies the client. Retrieve `user_id` from the context and pass it as the `uuid` param so the request matches the contract used by every other API call.

test(history): add unit tests for `pubnub_fetch_history` URL construction

Add six tests covering default options, all options enabled, custom user ID, auth key, in-progress guard, and 403 error handling. Each test verifies the full outgoing URL including the `uuid` parameter.


KGronek-Pubnub (commented):
If we use all possible parameters in `fetch_history`,  the last url parameter `auth` will be skipped - won't be encoded. I think you need to change PUBNUB_MAX_URL_PARAMS to 13 to make it work properly. 

Check: https://github.com/pubnub/c-core/pull/214 for reference.

KGronek-Pubnub (approved):
LGTM

parfeon:
@pubnub-release-bot release

PR #256, opened 2026-09-08
Use `const char*` for string-literal pointers

parfeon:
fix(cpp-compat): use `const char*` for string-literal pointers

Qualify `method` in `pubnub_publish_ex()` and `method_verb` in `pn_gen_pam_v3_sign()` as `const char*` so C-core compiles when `.c` files are built as C++.


KGronek-Pubnub (approved):
LGTM

parfeon:
@pubnub-release-bot release

PR #258, opened 2026-09-23
Prevent DNS server rotation from looping forever

parfeon:
fix(dns): prevent DNS server rotation from looping forever

Replace the compile-time `PUBNUB_MAX_DNS_SERVERS_MASK` exhaustion check with a `last_server_reached` flag set by `get_dns_ip`, and make `rotations_count` an unsigned field reset per transaction, so IPv4 resolution gives up instead of looping forever.

fix(sockets): guard IPv6 spare-address check in IPv4-only build

Guard the `AF_INET6` spare-address branch in `try_TCP_connect_spare_address` with `#if PUBNUB_USE_IPV6` so `PUBNUB_USE_IPV6=0` builds compile.

build(ci): verify IPv4-only socket transport build

Add a `run-tests.yml` job that compiles the socket transport with `PUBNUB_USE_IPV6=0` to keep the IPv4-only build from breaking again.

test(dns): cover DNS server rotation termination

Add `pubnub_dns_rotation_unit_test` over the real `get_dns_ip` and `pbpal_dns_rotate_server` to assert bounded termination and primary re-selection after a rotation.


KGronek-Pubnub (approved):
LGTM

parfeon:
@pubnub-release-bot release

